PR #26, opened 2020-09-30
add a sample ocp certificate policy, tuning still needed

gparvin:
This is a sample that monitors OCP certs.  OCP manages its own certs so this is really just a sample that uses the certificate policy controller.

### DIFF
--- a/community/SC-System-and-Communications-Protection/policy-ocp4-certs.yaml
+++ b/community/SC-System-and-Communications-Protection/policy-ocp4-certs.yaml
@@ -3,9 +3,9 @@ kind: Policy
 metadata:
   name: policy-cert-ocp4
   annotations:
-    policy.open-cluster-management.io/standards: NIST-CSF
-    policy.open-cluster-management.io/categories: PR.DS Data Security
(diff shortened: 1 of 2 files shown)


community/SC-System-and-Communications-Protection/policy-ocp4-certs.yaml
```diff
@@ -0,0 +1,85 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: policy-cert-ocp4
+  annotations:
+    policy.open-cluster-management.io/standards: NIST-CSF
```
ycao56: needs to be NIST 800-53 standards,categories and controls
should we put this in stable folders as these are officially supported?

gparvin: Right now it is just a sample on how to monitor the existing certificates in OCP.  I think it needs some more testing and likely will need to be modified based on which OCP CA is creating the certs (different rules may be needed for different namespaces)

ycao56 (approved):

dockerymick (approved):
Made a small change, description looks good !

PR #42, opened 2020-10-28
Provide policy that installs the official compliance operator

JAORMX:
This policy doesn't use the upstream release, but instead gets the
official and supported release from the red hat catalog.


dockerymick (approved):
Made a minor commit, changes look good to me

ycao56 (approved):

PR #61, opened 2020-12-10
add encryption status check policy template to etcd encryption policy

ycao56:


gparvin (approved):
Excellent!  I'll have Tsu-Phin take a look too so he's aware since he may have thought about this.

tphee (approved):

PR #65, opened 2020-12-17
Undeploy script

dhaiducek:
This undeploy script (AKA `remove.sh`) will look for matching `<prefix>-chan` Channel and `<prefix>-sub` Subscription resources and remove them in tandem.

Neither a namespace nor a prefix is required--the script will poll the cluster for both if not provided.

If the Subscription has been changed and no longer points to the Channel, the script refuses to delete the resource but gives the user a warning and the commands to do it themselves.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -24,7 +24,7 @@ Usage:
 
 # `remove.sh`
 
-## Remove resources from Red Hat Advanced Cluster Management created via `deploy.sh`
+## Find and remove resources from Red Hat Advanced Cluster Management created via `deploy.sh`
 ```
 Prerequisites:
  - oc CLI should be pointing to the cluster to remove from
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,11 +1,17 @@
-# `deploy.sh`
+# Deploy policies to Red Hat Advanced Cluster Management 
 
-## Deploy policies to Red Hat Advanced Cluster Management via GitOps
+Deploy policies to Red Hat Advanced Cluster Management with the `deploy.sh` script.
+
+## Deploying policies with GitOps
+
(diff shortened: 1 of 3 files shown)


ycao56:
adding @dockerymick to take a look at doc

ycao56:
Is `oc` a hard requirement for user to run this script? Can we do the same with just kubectl?

gparvin (approved):
Thanks for adding this enhancement!

dhaiducek:
Yes, `oc` is a hard requirement because it allows the script to find and iterate over the available namespaces. I'm not aware of a way to find available namespaces otherwise.

I changed `deploy.sh` to use `oc` also, but I can change it back to `kubectl` if it needs to be more generic. I was thinking the scripts would both be pointing at an Openshift RHACM hub, but maybe that's a poor assumption? If so, I can adjust the script to require a namespace be provided if a user doesn't have clusterwide access and move everything to `kubectl`.

ycao56:
I think it's not a problem for now. But in the future, after we open source the framework code, it should not require openshift to deploy and test. So it's better to make oc optional.

dhaiducek:
Ah, okay. Let me make that change now so that we don't have to worry about it later.

dockerymick (commented):
Made updates to the README @dhaiducek let me know what you think

dockerymick (approved):
Made an update for clarity

dhaiducek:
@dockerymick I made some additional adjustments to your edits. Most notably, I replaced "View the following example..." with "View the following guidance..." since to me it's not an example. And I also added a note that parameters are optional. What do you think?

ycao56 (approved):

dockerymick:
@dhaiducek great idea, i agree with your change

PR #82, opened 2021-02-05
Create policy report policy

gparvin:
Kyverno can be installed on managed clusters and you can create policies similar to a sample in progress now.  These policies create PolicyReport or ClusterPolicyReport CRs.  Inside these CRs are the results as to whether there was a policy failure.  To keep from having to look for these failures for each kyverno policy that is written, something like this generic policy report policy can be used to match any policy report failures.

Other tools could use the PolicyReport CRD also, in which case this policy will automatically detect those reports that have failures automatically.


gparvin:
too much got pull in somehow... closing and will create a cleaner pr

PR #83, opened 2021-02-05
create a policy that looks for policy report CRs that have failures

gparvin:
Kyverno can be installed on managed clusters and you can create policies similar to a sample in progress now. These policies create PolicyReport or ClusterPolicyReport CRs. Inside these CRs are the results as to whether there was a policy failure. To keep from having to look for these failures for each kyverno policy that is written, something like this generic policy report policy can be used to match any policy report failures.

Other tools could use the PolicyReport CRD also, in which case this policy will automatically detect those reports that have failures automatically.


ycao56 (approved):

community/README.md
```diff
@@ -84,6 +84,7 @@ Policy  | Description | Prerequisites
...
 [Integrity Shield](./CM-Configuration-Management/policy-integrity-shield.yaml)| Use the Integrity Shield to protect the integrity of Kubernetes resources in a cluster (e.g. OpenShift). | See the [Integrity Shield documentation.](https://github.com/open-cluster-management/integrity-shield/blob/master/docs/ACM/README_ENABLE_ISHIELD_PROTECTION_ACM_ENV.md)
 [Integrity Shield Events](./CM-Configuration-Management/policy-integrity-shield-events.yaml)| Use the Integrity Shield Events policy to show a status, which represents whether Integrity Shield has denied some requests in a cluster or not. | See the [Integrity Shield documentation.](https://github.com/open-cluster-management/integrity-shield/blob/master/docs/ACM/README_SHOW_INTEGRITY_STATUS_ON_ACM_UI.md)
+[Policy Report Failures](./CM-Configuration-Management/policy-check-reports.yaml)| The Policy Report Failures policy will search for any Policy Report resources that contain failures in the results. | An example of a tool that creates PolicyReports is [Kyverno](ihttps://kyverno.io/).  Other tools will likely create policy reports in the future.
```
dockerymick: I am a little confused with the name of the policy. For consistency to the policy name in Kyverno, and to help eliminate confusion, can the description be updated to the following:

PolicyReport failures | The `policy-check-reports` policy searches for any PolicyReport resources that contain failures in the results.

I would also remove the last sentence

community/CM-Configuration-Management/policy-check-reports.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: policy.open-cluster-management.io/v1
```
dockerymick, line 8: i would also add the description of the policy before the contents of the policy to help; this is not required, but can benefit the user

community/README.md
gparvin: @dockerymick I've updated the readme with your suggestions. Thanks for reviewing!

community/CM-Configuration-Management/policy-check-reports.yaml
gparvin: @dockerymick Take a look at what I just added and see if it looks ok.  Thanks!

dockerymick: @gparvin Thanks for your patience Gus. I suggest the following updates:

Set remediationAction to "inform" to receive a list of PolicyReport resources that display "Not compliant", when any PolicyReport resource has a status of "fail".

gparvin: @dockerymick Take a look at my update.  I reworded it some so want to make sure it's ok.

dockerymick: Looks good to me! @gparvin

dockerymick (approved):
Updates look good to me, thanks Gus :)

PR #89, opened 2021-02-23
add gatekeeper operator policy to stable folder

ycao56:


gparvin (approved):
This is great!  Thanks

stable/CM-Configuration-Management/policy-gatekeeper-operator-downstream.yaml
```diff
@@ -0,0 +1,90 @@
+# This policy verifies the installation of the official & supported version of
```
dockerymick: I would spell out "and" instead of using the symbol

dockerymick (approved):
Made one comment, otherwise this PR looks good to me!

PR #91, opened 2021-03-05
add finalizers to policy-integrity-shield for k8s v1.20 and later

hirokuni-kitahara:
- On K8s v1.20 and later, kubernetes garbage collector will ignore cluster scope resources that have non-existing ownerReference. So Integrity Shield uses finalizer to delete its cluster scope operands.
- Also confirmed it can work on K8s 1.19 or earlier.


hirokuni-kitahara:
@gparvin Could you review this PR? Thank you.
/assign @gparvin

gparvin (approved):

PR #96, opened 2021-03-22
ACS sensor policy

gparvin:
This policy can be used to alert whether the managed clusters have the sensor installed and configured for the Stackrox central server that manages the clusters that also managed by ACM.  It can help make sure there's management consistency across the clusters.


gparvin:
Something isn't write with my repo -- will have to delete and recreate.

PR #101, opened 2021-03-30
Add policy for OpenShift CIS via the Compliance Operator

JAORMX:
This adds a policy that ensures a scan exists for the OpenShift CIS
benchmark. This is executed by the Compliance Operator, and thus it
requires that the installation policy is used as well.

Signed-off-by: Juan Antonio Osorio Robles <jaosorior@redhat.com>


dockerymick (approved):
I made small updates to both files. These additions look great, thank you @JAORMX

JAORMX:
@dockerymick thanks for the corrections!

PR #110, opened 2021-05-10
this appears to be an extra entry

gparvin:
I don't see anything missing.  I think this entry just isn't needed.


ycao56:
@gparvin I think this should be right link https://github.com/open-cluster-management/policy-collection/blob/main/community/CM-Configuration-Management/policy-github-oauth-sample.yaml

ycao56:
can you fix it?

ycao56:
nvm you are right. this is not needed

petr-muller:
@ycao56 @gparvin This PR hit a GH limit on the number of statuses on a SHA and Prow now cannot merge it. The reason for this is that you made it Tide-mergeable with slash-commands, but your repo is at the same time configured to only allow merges when a PR has an **approving GH review**.

In order for this PR to merge, it needs to be amended (so that it gets a new SHA) and then approved via GH review. Long-term, I suggest to disable the custom merge criteria b/c you have Tide configured.

ycao56 (approved):

PR #118, opened 2021-06-07
create a policyreport policy for the v1alpha2 api

gparvin:
Create a PolicyReports policy for the new api.  Same as the policy-check-reports.yaml policy except it uses the new API.


ycao56:
should we consider removing the policy using old api?

dockerymick (approved):
Made a couple of small updates, but this looks good to me! Thanks for pulling me in

PR #143, opened 2021-08-19
added policies

ch-stark:
Signed-off-by: Christian Stark <cstark@cstark.muc.csb>


community/CM-Configuration-Management/policy-ansible-awx-operator.yaml
```diff
@@ -0,0 +1,69 @@
...
+                  name: awx-resource-operator
+                  namespace: policies
+                spec:
+                  channel: release-0.1
+                  installPlanApproval: Automatic
+                  name: awx-resource-operator
+                  source: redhat-operators
+                  sourceNamespace: openshift-marketplace
+                  startingCSV: awx-resource-operator.v0.1.1
+---
```
gparvin, line 58: A similar policy didn't work for me... make sure this operator installs ok.  I think it needs an OperatorGroup.

community/CM-Configuration-Management/policy-cluster-network-sample.yaml
```diff
@@ -58,3 +58,21 @@ subjects:
 - name: policy-config-network
   kind: Policy
   apiGroup: policy.open-cluster-management.io
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
+metadata:
+  name: placement-policy-config-network
+  namespace: policies
```
gparvin: The namespace shouldn't be hard coded here.  It needs to be removed.

community/CM-Configuration-Management/policy-custom-catalog.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
```
gparvin, line 3: The `name` is missing.

community/CM-Configuration-Management/policy-kyverno.yaml
```diff
@@ -0,0 +1,142 @@
...
+subjects:
+  - name: policy-install-kyverno
+    kind: Policy
+    apiGroup: policy.open-cluster-management.io
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
+metadata:
+  name: placement-policy-install-kyverno
+  namespace: kyverno
```
gparvin: The namespace shouldn't be in the placement rule.

gparvin (commented):
Added a few comments.  Thanks for working on this!

community/CM-Configuration-Management/policy-cluster-network-sample.yaml
ch-stark: thanks, good catch

community/CM-Configuration-Management/policy-custom-catalog.yaml
ch-stark: thanks

community/CM-Configuration-Management/policy-kyverno.yaml
ch-stark: can we discuss if I can leave it that way?

gparvin: The Policy and it's associated PlacementRule and PlacementBinding all need to be in the same namespace.  We want that namespace to be selected by the customer.
Note that you do have another PlacementRule embedded in the policy in the kyverno namespace.  That one is ok to be in the kyverno namespace, but this one needs to match the root namespace selected by the customer for the policy.

gparvin (commented):
Double check the kyverno policy and see if the comment I added makes sense.

ch-stark:
@gparvin thanks I removed the namespace from placement-binding/rule

gparvin (approved):
Thank you for all these great policies!

PR #148, opened 2021-09-13
Added gatekeeper policy that disallows anonymous users in the cluster

michaelkotelnikov:
The [NSA / CISA Kubernetes Hardening Guidance (August 2021)](https://media.defense.gov/2021/Aug/03/2002820425/-1/-1/1/CTR_KUBERNETES%20HARDENING%20GUIDANCE.PDF) [page 19] states that -

"
In Kubernetes 1.6 and newer, anonymous requests are enabled by default. When RBAC is enabled, anonymous requests require explicit authorization of the system:anonymous user or system:unauthenticated group. Anonymous requests should be disabled by passing the --anonymous-auth=false option to the API server. Leaving anonymous requests enabled could allow a cyber actor to access cluster resources without authentication.
"

The recommendation states that anonymous authentication should be disabled by default. Yet, many environments leave `--anonymous-auth` enabled for various reasons. The purpose of this policy is to control and reduce the usage of anonymous requests in Kubernetes environments.

***Scenario 1 - Policy is not deployed*** - 

```
# as user-1, local namespace admin -

$ cat disallowed-binding.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: dangerous-rolebinding
  namespace: application-1
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:unauthenticated

$ kubectl apply -f /tmp/disallowed-binding.yaml 
rolebinding.rbac.authorization.k8s.io/dangerous-rolebinding created

$ kubectl get pods --as="system:anonymous" -n application-1
NAME                         READY   STATUS    RESTARTS   AGE
webserver-7cccc68cc6-dmkt5   1/1     Running   0          8m28s
```

***Scenario 2 - Policy is deployed*** -

```
# as user-1, local namespace admin -

$ kubectl apply -f disallowed-binding.yaml 
Error from server ([denied by no-anonymous] Unauthenticated user reference is not allowed in RoleBinding dangerous-rolebinding ): error when creating "/tmp/disallowed-binding.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [denied by no-anonymous] Unauthenticated user reference is not allowed in RoleBinding dangerous-rolebinding
```

Note that the policy is set to `enforcementAction` is set to `dryrun` - the Gatekeeper policy is not enforced by default.

Feel free to share feedback or edit the contents of the PR :)

### DIFF
--- a/community/AC-Access-Control/policy-gatekeeper-disallow-anonymous.yaml
+++ b/community/AC-Access-Control/policy-gatekeeper-disallow-anonymous.yaml
@@ -1,0 +1,126 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: policy-gatekeeper-disallow-anonymous
+  annotations:
+    policy.open-cluster-management.io/standards: NIST SP 800-53
+    policy.open-cluster-management.io/categories: AC Access Control
+    policy.open-cluster-management.io/controls: AC-2 Account Management
+spec:
+  remediationAction: enforce 
+  disabled: false
+  policy-templates:
+    - objectDefinition:
+        apiVersion: policy.open-cluster-management.io/v1
+        kind: ConfigurationPolicy
+        metadata:
+          name: policy-gatekeeper-disallow-anonymous
+        spec:
+          remediationAction: enforce
+          severity: low
+          object-templates: 
+            - complianceType: musthave
+              objectDefinition:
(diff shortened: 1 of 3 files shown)


michaelkotelnikov:
/assign @chunxialexluo

FYI @ch-stark

PR #150, opened 2021-09-15
remove check for install config that isn't working

gparvin:
Fix the fips policy. See issue: https://github.com/open-cluster-management/backlog/issues/15947

@ch-stark The install-config.yaml check doesn't work.  I thought we had checked that it did work so I'm not sure if we have a regression there.  To me it seems like the data in the configmap may have changed into one big string -- though I'm not sure if that wasn't the case before.

### DIFF
--- a/community/SC-System-and-Communications-Protection/policy-check-fips.yaml
+++ b/community/SC-System-and-Communications-Protection/policy-check-fips.yaml
@@ -18,11 +18,6 @@ spec:
       spec:
         remediationAction: inform
         severity: low
-        namespaceSelector:
-          exclude:
-            - kube-*
-          include:
-            - default
         object-templates:
           - complianceType: musthave
             objectDefinition:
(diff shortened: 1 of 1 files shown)


gparvin:
/hold
Investigating issue with the fips policy.

ch-stark:
it's correct, look, if you ONLY need FIPS, a check for this MC object is enough oc get mc 99-master-fips -o yaml

Thanks @gparvin 

apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  creationTimestamp: "2021-05-19T05:07:20Z"
  generation: 1
  labels:
    machineconfiguration.openshift.io/role: master
  name: 99-master-fips
  resourceVersion: "1627"
  uid: bb42add2-2d8d-44c3-b78b-6b37c0f69487
spec:
  config:
    ignition:
      version: 3.2.0
  extensions: null
  fips: true
  kernelArguments: null
  kernelType: ""
  osImageURL: ""

gparvin:
This PR is wrong.  I'll create a new one.

PR #164, opened 2021-10-15
add details about being a subscription admin

gparvin:
We need to say something about being a subscription admin.  I added an update to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,9 @@ applied automatically.
 ### Subscription Administrator
 
 In new versions of Open Cluster Management you must be a subscription administrator in order to deploy
-policies using a subscription.  In these cases the subscription is still successfully created, but policy
-resources are not distributed as expected.  You can view the status of the subscription to see the subscription
-errors.  If subscription admin is required, you will see a message similar to the one below for any resource that
-was not created:
+policies using a subscription. In these cases the subscription is still successfully created, but policy
+resources are not distributed as expected. You can view the status of the subscription to see the subscription
+errors. If the subscription administrator role is required, a message similar to the following one appears for any resource that is not created:
 
 ```
         demo-stable-policies-chan-Policy-policy-cert-ocp4:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,10 +41,10 @@ applied automatically.
 
 ### Subscription Administrator
 
-In new version of Open Cluster Management you must be a subscription administrator in order to deploy
+In new versions of Open Cluster Management you must be a subscription administrator in order to deploy
 policies using a subscription.  In these cases the subscription is still successfully created, but policy
 resources are not distributed as expected.  You can view the status of the subscription to see the subscription
-errors.  If subscription admin is required, you will see a message similar to this for any resource that
+errors.  If subscription admin is required, you will see a message similar to the one below for any resource that
```
dockerymick: There is an extra space before "If". "was not created" is also in a new line. I suggest the following update:

If the subscription administrator role is required, a message similar to the following one appears for any resource that is not created:

README.md
```diff
@@ -54,7 +54,8 @@ was not created:
           reason: 'not deployed by a subscription admin. the resource apiVersion: policy.open-cluster-management.io/v1 kind: Policy is not deployed'
 ```
 
-To become a subscription admin, you must add an entry for your user to the `open-cluster-management:subscription-admin` ClusterRoleBinding.
+To become a subscription administrator, you must add an entry for your user to
```
dockerymick: I suggest to move "the `open-cluster-management:subscription-admin` ClusterRoleBinding." to line 57. i would also monospace, so change ClusterRoleBinding to `ClusterRoleBinding`.

dockerymick (changes requested):

README.md
gparvin: Thanks, I made this update.

README.md
gparvin: I made this update.  Take a look when you get a chance. @dockerymick

dockerymick (approved):
looks good to me, thanks Gus!

PR #196, opened 2021-11-29
To use AAP with ACM we need to not have the targetNamespace set

gparvin:
The targetNamespace causes the CSV to be deployed for AAP in namespace
mode instead of in cluster mode.  We need cluster mode so the
controllers will listen to AnsibleJob resource creations across all
namespaces.


ycao56 (approved):

gparvin:
This should fix issue https://github.com/open-cluster-management/backlog/issues/18165

PR #198, opened 2021-11-29
Switch gatekeeper policies to use dryrun enforcement

gparvin:
To make the gatekeeper policies "inform" by default, they need to use
the dryrun `enforcementAction`.  This way you can learn the impacts to
your cluster before removing the `dryrun` and requiring enforcement.

Refs:
 - https://github.com/open-cluster-management/policy-collection/issues/127


gparvin:
Should we mention that this is using `dryrun` in the README?

ycao56 (approved):

PR #202, opened 2021-12-16
Add additional checks to detect if operator is running

ycao56:
Add additional checks to detect if operator is running

resolves https://github.com/open-cluster-management/backlog/issues/18566

Signed-off-by: Yu Cao <ycao@redhat.com>

### DIFF
--- a/community/CM-Configuration-Management/policy-automation-operator.yaml
+++ b/community/CM-Configuration-Management/policy-automation-operator.yaml
@@ -17,10 +17,10 @@ spec:
         apiVersion: policy.open-cluster-management.io/v1
         kind: ConfigurationPolicy
         metadata:
-          name: ansible-automation-platform-ns
+          name: ansible-automation-platform
         spec:
           remediationAction: inform
-          severity: low
+          severity: high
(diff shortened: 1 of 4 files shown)


ycao56:
@jnpacker @ch-stark FYI

dhaiducek (approved):
This looks good!

Is the CSV the most reliable check you found? I was thinking from a generator perspective, it would be simpler to generate a `status` check for a resource that was already provided (otherwise the user is pretty much writing the entire policy). Maybe something like:
```yaml
- complianceType: mustnothave
  objectDefinition:
    <reuse metadata from a given resource, like the subscription>
    status:
      catalogHealth:
      - healthy: false
```
I'll look into it for the generator, but from what you've seen would something like that work to check the health of the Subscription or does that not really report on the resources it backs?

ycao56:
@dhaiducek Yeah, I don't think checking subscription status field is enough. I found CSV is more reliable based on my testing. If for some reason the pods are not installed an running, CSV would return phase `Failed` or `Installing`. Subscription couldn't show that granularity.

PR #207, opened 2021-12-23
Revert "Revert "Add additional checks to detect if operator is running (#202)""

dhaiducek:
Reverts open-cluster-management/policy-collection#206

The original goal in reverting the policy changes was to fix tests in the canaries (tests that had also been updated). However, the tests hadn't actually arrived in the canaries when the change was reverted, so after these changes were reverted the canaries also failed. Restoring the original change should make all tests pass again.


mprahl:
@dhaiducek could you please provide a little more context in the PR description?

dhaiducek:
Description updated

ycao56 (approved):

mprahl (approved):

PR #208, opened 2022-01-05
Organization rename to `stolostron`

dhaiducek:
- https://github.com/stolostron/backlog/issues/18778

The only caveat here is that Integrity Shield wasn't migrated to the new org in Quay, so that policy may need to be updated if/when it migrates.


CODE_OF_CONDUCT.md
```diff
@@ -1,3 +1,3 @@
 # Code of Conduct
 
-Refer to our [Open Cluster Management Community Code of Conduct](https://github.com/open-cluster-management/community/blob/main/CODE_OF_CONDUCT.md)
\ No newline at end of file
+Refer to our [Open Cluster Management Community Code of Conduct](https://github.com/stolostron/community/blob/main/CODE_OF_CONDUCT.md)
```
ycao56: https://github.com/open-cluster-management/community/blob/main/CODE_OF_CONDUCT.md redirects to `-io` at the moment.

dhaiducek: Wow--you're right. The `community` repo in `stolostron` doesn't even exist. I'll update it.

ycao56 (approved):

PR #210, opened 2022-01-07
Updating the link to psp policy yaml

dockerymick:
Signed-off-by: Mikela Jackson <mdockery@redhat.com>

### DIFF
--- a/community/README.md
+++ b/community/README.md
@@ -195,7 +195,7 @@ No policies yet       |  |
 Policy  | Description | Prerequisites
 ------- | ----------- | -------------
 [OpenShift Certificate Expiration Policy](./SC-System-and-Communications-Protection/policy-ocp4-certs.yaml) | Monitor the OpenShift 4.x namespaces to validate that certificates managed by the infrastructure are rotated as expected. | OpenShift 4.x is required.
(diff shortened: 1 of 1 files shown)


gparvin (commented):
I think we should delete this entry in the community readme, since it looks like we are pointing to the stable policy.  The note here should get copied to the stable readme.  That way we won't have this policy listed in 2 readmes with different info.

dockerymick:
@gparvin Thanks for your feedback ! Making sure that I understood your suggestion, you are suggesting that the Pod Security Policy entry should be removed from the community/README only? I saw that it is linked correctly in the [stable README](https://github.com/stolostron/policy-collection/tree/main/stable#system-and-communications-protection).

gparvin:
@dockerymick Yes, since it is already linked correctly in the stable README, it should not be listed at all in the community readme, it just needs to be removed.  I see that we deprecated this policy -- so I suspect there may have been some movement of this policy that caused some confusion.  I'll open a doc issue because I see a bad link there too.

dockerymick:
@gparvin Sounds good! Thanks for opening a doc issue

ycao56 (approved):

PR #226, opened 2022-02-10
Update subscription-admin instructions and policy

dhaiducek:
I found that when deploying a Subscription as `kube:admin`, the user is `system:admin`, so the Subscription never deploys, even with a binding to `kube:admin`.

Also clarify our docs that the Subscription will not self-reconcile when user permissions change. The Subscription needs to be redeployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,7 @@ subjects:
     name: my-username
 ```
 
-After updating the `ClusterRoleBinding`, you can delete the subscription and run the deploy script
-again or simply wait for the subscription controller to synchronize the resources.
+After updating the `ClusterRoleBinding`, you need to delete the subscription and deploy the subscription again.
 
 ### Policy Generator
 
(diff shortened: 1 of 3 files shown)


ckandag:
@dhaiducek : How did you know the kube:admin user is internally called system:admin ?  was it being logged in the SubscriptionAdmin error msgs  that  user `system:admin` does not have access ?

dhaiducek:
Yes--the Subscription has a `user-identity` annotation that can be decoded to the user that deployed it. When I deployed it with the CLI, the user was `system:admin`. (Plus, when you do `oc whoami`, the result is `system:admin`.) I'm not sure why the CLI is different from the UI with `kube:admin` in Kubernetes, but that's what I found when I tried it out and it's also in the logs (See https://github.com/stolostron/backlog/issues/19816#issuecomment-1034262677 for some additional details)

ckandag:
the policy appears to already have a username set it in so Approving.

But In general, I think the community policies just have a place-holders  for usernames - so the policy user can set it as appropriate.

ckandag (approved):

PR #229, opened 2022-02-16
Fix problems with ODF and eliminate need for scripts

gparvin:
There were some problems with the ODF install.  The operator now
installs and the storagesystem and the storagecluster are properly
setup.  Observability I think is still have trouble writing to noobaa,
but there are statistics and details that show things are closer to
working properly.  While I made those changes I also eliminated the
scripts required by taking Andrew's advice and incorporating the post
scripts as jobs that run in the local-quay and stackrox namespaces
to setup secrets those apps need.


ycao56 (approved):

gparvin:
Issue resolved is: https://github.com/stolostron/backlog/issues/20044

PR #233, opened 2022-02-25
Specify the remediationAction in each template instead

ycao56:
Make it more readable and less confusing as the second template doesn't support enfoce
Signed-off-by: Yu Cao <ycao@redhat.com>


gparvin (approved):
Thanks

PR #235, opened 2022-03-01
Update README.md

dockerymick:
Signed-off-by: Mikela Jackson <mzjackson@redhat.com>


gparvin (commented):
Looks good.  I'm not sure if this one should replace the older Integrity Shield blog entry.  @jrnp7 may know.  It may be nice to group them.

jrnp7:
This will not replace the previous one but agree it would be good to group all the three Integrity Shield blogs @gparvin @dockerymick

gparvin (approved):
@dockerymick I see these are together in the list of blogs.  I think this is good -- we really have enough blogs on this one topic now we may need to add more details to recommend to customers to "start here".

PR #239, opened 2022-03-03
Added an example policy

hellfire114:


hellfire114 (commented):
Signed-off-by: Raghu <rbanda@redhat.com>

PR #240, opened 2022-03-03
Added example policy for managed clusters to have viewonly and admin type of roles across groups

hellfire114:


gparvin (commented):
Thanks for creating this policy and contributing it.  We would like a README.md update to help provide similar information you've added already in the comments.  The README.md file is one directory up in the community folder.

gparvin (dismissed):
Thanks for your contribution!

PR #244, opened 2022-03-10
test-IAM-policy-in-policy-generator

ChunxiAlexLuo:
Signed-off-by: Chunxi Luo <chuluo@redhat.com>

@gparvin 

1) So as this PR, if we just put `objectDefinition` section in the policy manifest `policy-limitclusteradmin.yaml` and move the rest higher-level info into `policyGenerator.yaml`, then we got the below result in the output:

```
kind: Policy
metadata:
  annotations:
    policy.open-cluster-management.io/categories: AC Access Control
    policy.open-cluster-management.io/controls: AC-3 Access Enforcement
    policy.open-cluster-management.io/standards: NIST SP 800-53
  labels:
    open-cluster-management.io/policy-set: openshift-hardening
  name: policy-limitclusteradmin
  namespace: policies
spec:
  disabled: false
  policy-templates:
  - objectDefinition:
      kind: ConfigurationPolicy
      metadata:
        name: policy-limitclusteradmin
      spec:
        object-templates:
        - complianceType: musthave
          objectDefinition:
            kind: IamPolicy
            metadata:
              name: policy-limitclusteradmin-example
            spec:
              maxClusterRoleBindingUsers: 5
              namespaceSelector:
                exclude:
                - kube-*
                - openshift-*
                include:
                - '*'
              remediationAction: inform
              severity: medium
        remediationAction: inform
        severity: medium
```
2) Current policy-generator + policy-collection will get the below result in the output, we can find policy-generator just treat the policy manifest `policy-limitclusteradmin.yaml`  as an `objectDefinition` section, that is the reason of nested `policy-templates`. If we need to keep the policy manifest `policy-limitclusteradmin.yaml`  as current format, then we should add an additional boolean flag in the policy section of `policyGenerator.yaml` to let policy-generator subtract `objectDefinition` part from the policy manifest, rather than read the whole manifest as an `objectDefinition` section.

Feel free to let me know which direction we should go.

```
kind: Policy
metadata:
  annotations:
    policy.open-cluster-management.io/categories: CM Configuration Management
    policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
    policy.open-cluster-management.io/standards: NIST SP 800-53
  labels:
    open-cluster-management.io/policy-set: openshift-hardening
  name: policy-limitclusteradmin
  namespace: policies
spec:
  disabled: false
  policy-templates:
  - objectDefinition:
      kind: ConfigurationPolicy
      metadata:
        name: policy-limitclusteradmin
      spec:
        object-templates:
        - complianceType: musthave
          objectDefinition:
            kind: Policy
            metadata:
              annotations:
                policy.open-cluster-management.io/categories: AC Access Control
                policy.open-cluster-management.io/controls: AC-3 Access Enforcement
                policy.open-cluster-management.io/standards: NIST SP 800-53
              name: policy-limitclusteradmin
            spec:
              disabled: false
              policy-templates:
              - objectDefinition:
                  kind: IamPolicy
                  metadata:
                    name: policy-limitclusteradmin-example
                  spec:
                    maxClusterRoleBindingUsers: 5
                    namespaceSelector:
                      exclude:
                      - kube-*
                      - openshift-*
                      include:
                      - '*'
                    remediationAction: inform
                    severity: medium
              remediationAction: inform
        remediationAction: enforce
        severity: medium
```

### DIFF
--- a/policygenerator/policy-sets/openshift-hardening/input-admin/policy-limitclusteradmin.yaml
+++ b/policygenerator/policy-sets/openshift-hardening/input-admin/policy-limitclusteradmin.yaml
@@ -1,24 +1,11 @@
 apiVersion: policy.open-cluster-management.io/v1
-kind: Policy
+kind: IamPolicy # limit clusteradminrole and report violation
 metadata:
-  name: policy-limitclusteradmin
-  annotations:
-    policy.open-cluster-management.io/standards: NIST SP 800-53
-    policy.open-cluster-management.io/categories: AC Access Control
-    policy.open-cluster-management.io/controls: AC-3 Access Enforcement
+  name: policy-limitclusteradmin-example
 spec:
-  remediationAction: inform
-  disabled: false
-  policy-templates:
-    - objectDefinition:
-        apiVersion: policy.open-cluster-management.io/v1
-        kind: IamPolicy # limit clusteradminrole and report violation
-        metadata:
-          name: policy-limitclusteradmin-example
-        spec:
-          severity: medium
-          namespaceSelector:
-            include: ["*"]
-            exclude: ["kube-*", "openshift-*"]
-          remediationAction: inform # the policy-template spec.remediationAction is overridden by the preceding parameter value for spec.remediationAction.
-          maxClusterRoleBindingUsers: 5
+  severity: medium
+  namespaceSelector:
+    include: ["*"]
+    exclude: ["kube-*", "openshift-*"]
+  remediationAction: inform # the policy-template spec.remediationAction is overridden by the preceding parameter value for spec.remediationAction.
+  maxClusterRoleBindingUsers: 5
--- a/policygenerator/policy-sets/openshift-hardening/policyGenerator.yaml
+++ b/policygenerator/policy-sets/openshift-hardening/policyGenerator.yaml
@@ -17,9 +17,16 @@ policyDefaults:
   standards:
     - NIST SP 800-53
 policies:
-#- name: policy-limitclusteradmin # iam policy not supported by the generator
-#  manifests:
-#    - path: input-admin/policy-limitclusteradmin.yaml
+- name: policy-limitclusteradmin # iam policy not supported by the generator
+  standards: 
+    - NIST SP 800-53
+  categories: 
+    - AC Access Control
+  controls: 
+    - AC-3 Access Enforcement
+  remediationAction: inform
+  manifests:
+    - path: input-admin/policy-limitclusteradmin.yaml
 - name: policy-remove-kubeadmin
   categories:
     - SC System and Communications Protection


dhaiducek:
Having the `config-policy-controller` deploy an `IamPolicy` is convoluted (and probably wouldn't return the compliance status all the way back to the hub). The idea of the enhancement is to support policy types other than `ConfigurationPolicy`--this should be addressed by updating the generator code to support it.

dhaiducek:
Sorry if I'm misunderstanding your explanations. To me, I think there are a couple options that come to mind:
- When a policy is given through `policies[].manifests[].path` in the generator manifest, if it's a Policy kind, extract the `objectDefinition` object and dump it into the policy (and probably validate first to make sure it's one of ConfigurationPolicy, IamPolicy, CertPolicy)
- Create a new `policies[].path` that can validate and use a Policy kind manifest as-is

I'm not sure there's much desire in allowing users to have an IamPolicy or CertPolicy manifest since I'm assuming those would probably already be wrapped in a Policy.

dhaiducek:
@mprahl @ycao56 Do you have any further feedback on this?

ycao56:
> Sorry if I'm misunderstanding your explanations. To me, I think there are a couple options that come to mind:
> 
>     * When a policy is given through `policies[].manifests[].path` in the generator manifest, if it's a Policy kind, extract the `objectDefinition` object and dump it into the policy (and probably validate first to make sure it's one of ConfigurationPolicy, IamPolicy, CertPolicy)
> 
>     * Create a new `policies[].path` that can validate and use a Policy kind manifest as-is
> 
> 
> I'm not sure there's much desire in allowing users to have an IamPolicy or CertPolicy manifest since I'm assuming those would probably already be wrapped in a Policy.

Both makes sense. The second approach might have higher priority if considering we are going to merge iam and cert policy into config policy. 
Agree? @gparvin @dhaiducek @mprahl

mprahl:
My preference would be to have a new field under `manifests` called `policyPath` which can point to a YAML file of a non-root policy (e.g. ConfigurationPolicy, IAMPolicy, CertPolicy). Then the non-root policy is taken as is and added to the `policy-templates` array. There would then be validation to ensure that `manifest[0].path` and `manifest[0].policyPath` are not both set.

I believe this makes the user experience fairly consistent with the existing user experience.

dhaiducek:
I agree, @mprahl. I guess the question is: do we want to enable users to take an existing parent Policy manifest and use it as-is through the generator? The policy-collection has been based on those (and customers probably have them also at this point). Then that policy can be used in a set, modify placement, etc (though it's not really being "generated" any more).

mprahl:
> I agree, @mprahl. I guess the question is: do we want to enable users to take an existing parent Policy manifest and use it as-is through the generator? The policy-collection has been based on those (and customers probably have them also at this point). Then that policy can be used in a set, modify placement, etc (though it's not really being "generated" any more).

It's a good point. I think we don't want to support that since the Policy Generator can already generate policy sets that contains policies that are not generated by the Policy Generator. In my opinion, any root policy not generated by the Policy Generator should be deployed outside of the Policy Generator.

dhaiducek:
> the Policy Generator can already generate policy sets that contains policies that are not generated by the Policy Generator

It can? I think I missed that...

dhaiducek:
Oh! You just provide the name of the policy. Sorry--yes, in that case I 100% agree.

ChunxiAlexLuo:
Thanks for the clarification. Close this issue and open new PR based on feedback.

ChunxiAlexLuo:
@dhaiducek @mprahl @gparvin  

So for Gus user case here 
https://github.com/stolostron/backlog/issues/19466#issuecomment-1047886710

https://github.com/stolostron/policy-collection/blob/main/policygenerator/policy-sets/community/openshift-hardening/input-admin/policy-limitclusteradmin.yaml

It seems like root policy which should be directly added here 
https://github.com/stolostron/policy-generator-plugin/blob/7f358d41b64d2295b44756dd11d27cf8dff828e4/internal/plugin.go#L839
rather than put into `policyTemplates`

Just confirm we will not support this case.

ChunxiAlexLuo:
closed as we have decided to not support the root/full policy in the generator

PR #245, opened 2022-03-15
Implement `--dry-run` and `--deploy-app` flags for `deploy.sh`

dhaiducek:
- `deploy.sh` outputs the resulting Subscription manifests to a file every time, regardless of whether in dry-run mode or not
- `--dry-run` - Output the result to `stdout` but don't apply it to the cluster (also saves the result to `manifests.yaml`)
- `--deploy-app` - Deploy an Application manifest for additional visibility in the UI (also updates the Subscription to use Placement since the UI doesn't support the full topology otherwise)
- Updates to the various `README.md` for these changes as well as clarifications around the generator and PolicySets

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -9,10 +9,10 @@ help () {
   echo ""
   echo "Prerequisites:"
   echo " - oc or kubectl CLI must be pointing to the cluster to which to deploy policies"
-  echo " - The desired cluster namespace should already exist"
   echo ""
   echo "Usage:"
-  echo "  ./deploy.sh [-u <url>] [-b <branch>] [-p <path/to/dir>] [-n <namespace>] [-a|--name <resource-name>]"
+  echo "  ./deploy.sh [-u <url>] [-b <branch>] [-p <path/to/dir>] [-n <namespace>]"
+  echo "              [-a|--name <resource-name>] [--deploy-app] [--dry-run]"
   echo ""
   echo "  -h|--help                   Display this menu"
   echo "  -u|--url <url>              URL to the Git repository"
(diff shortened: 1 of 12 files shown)


deploy/deploy.sh
```diff
@@ -98,6 +110,20 @@ while read -r -p "Would you like to proceed (y/n)? " response; do
   esac
 done
 
+# Check for the namespace
```
gparvin: Should the namespace be created in a dry run scenario?

gparvin (changes requested):

deploy/deploy.sh
dhaiducek: Ooo, you're right--good catch!

gparvin:
/lgtm
This looks like great stuff!  Thanks for the updates.

dhaiducek:
/hold I'm not 100% sure we should add the Application feature: https://coreos.slack.com/archives/C027TN14SGJ/p1647613171020129?thread_ts=1647612794.419359&cid=C027TN14SGJ

(Well, at least it looks like we should add a Placement[Rule] if we do go down that road)

dhaiducek:
Okay, PlacementRule added. (Which we'll have to update to Placement eventually, but that might be a task for another day...) Please take another look. The logic flows got a little bit more complex with Placement in the mix, but I'm not sure about alternative tooling (or Kustomize features I may not be using).

gparvin:
Ah, yes, so you want it to look nice in the Topology part of App Lifecycle.  It will be a nice thing to show off in the sync call too.
/lgtm

gparvin (approved):

PR #254, opened 2022-04-06
Move the kyverno generate policies to the stable folder

gparvin:
Kyverno can help you generate resources that are required for your
compliance needs.  Here are 3 examples that let you copy secrets,
create initial NetworkPolicies and setup resource limits.  These
policies are being moved from community to stable as part of an
effort to highlight kyverno and its capabilities with open cluster
management.

Refs:
 - https://github.com/stolostron/backlog/issues/20388


mprahl:
@gparvin do we have tests for these policies?

gparvin:
> @gparvin do we have tests for these policies?

Not yet.  They need to be created.

mprahl (approved):

PR #258, opened 2022-04-19
Update the owners file to reflect those needing access

gparvin:
Remove and add members that need or don't need access.

Refs:
 - https://github.com/stolostron/backlog/issues/21546


ChunxiAlexLuo (approved):
/lgtm

PR #259, opened 2022-04-21
Container security operator display name changed

gparvin:
The container security operator now has a new display name.
This is causing problems with the CSV check in the automation.


ChunxiAlexLuo (approved):
/lgtm

/hold

PR #263, opened 2022-04-29
The new ACS operator wants to be installed in the rhacs-operator ns

gparvin:
Adjust the policy so the ACS operator is installed in the namespace
that it wants to be installed in now: `rhacs-operator`.  This creates
the new namespace and the operator group.

Refs:
 - https://github.com/stolostron/policy-collection/issues/260


ChunxiAlexLuo (approved):
/lgtm

All YAML update looks good to me

PR #264, opened 2022-05-02
add serverless policy

willkutler:
Signed-off-by: Will Kutler <wkutler@redhat.com>

https://github.com/stolostron/policy-collection/issues/243

not sure if the category/folder is correct, let me know if it needs to be changed

### DIFF
--- a/community/CM-Configuration-Management/policy-openshift-serverless.yaml
+++ b/community/CM-Configuration-Management/policy-openshift-serverless.yaml
@@ -1,0 +1,118 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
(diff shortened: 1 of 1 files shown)


gparvin (commented):
Our convention is to make policies "inform" by default in the community.  Can you also add an entry in the README.md for this policy?

gparvin:
Another user I think just donated this policy.  @willkutler Can you take a look at https://github.com/stolostron/policy-collection/blob/main/community/CM-Configuration-Management/policy-openshift-serverless.yaml and see if there's anything you think this PR could still be used for?  Otherwise I think it can be closed.  Thanks!

gparvin:
I think this was delivered in a different pr.  Please re-open if necessary.

PR #285, opened 2022-06-07
Pin the version of the community gatekeeper

JustinKuli:
The latest tag is sometimes updated to versions that may not work. This
pins the version to a currently working tag.

Refs:
 - https://github.com/stolostron/backlog/issues/23005

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>


dhaiducek (approved):
Good idea. I think this matches other policies in the collection.

One aside is that we should probably start maintaining a "things we should update each release" list since I could see us forgetting to update these regularly.

gparvin (approved):

PR #288, opened 2022-06-15
Update policy to point to the 0.4 volsync release

tesshuflower:
- This matches what we deploy with ACM 2.5 and also we'd like
  any users on older ACM 2.4 release to use this version as well.

Signed-off-by: Tesshu Flower <tflower@redhat.com>


dhaiducek (approved):
Thanks for the update, @tesshuflower!

PR #292, opened 2022-06-24
Observability was not being enabled properly due to bad config

gparvin:
Configuration that I believe previously "worked" was no longer working.
These changes fix the problem at least for my local test cluster.  I'll
need to setup a new cluster to make sure these updates aren't cluster
specific, but Observability is working great at the moment.


ChunxiAlexLuo (approved):
/lgtm (quay update)

PR #300, opened 2022-07-11
fix #299 - properly select namespaces

### LINKED ISSUE #299: kubeadmin removal policy has wrong namespaceSelector
In https://github.com/stolostron/policy-collection/blob/ad66e6b90a83f6f98d044596b42bc0c436d4a879/community/SC-System-and-Communications-Protection/policy-remove-kubeadmin.yaml

We have a:

```
          namespaceSelector:
            exclude:
              - kube-*
            include:
              - default
```

And at the same time want that the kubeadmin Secret in `kube-system` is removed. But kube-system is excluded through the selector.

duritong:
The present namespace selector does not make sense, since it
excludes `kube-*` while the Secret in question is in `kube-system`

Also the objectDefinition requires the `apiVersion` to be present
otherwise we do not lookup the proper kind.

### DIFF
--- a/community/SC-System-and-Communications-Protection/policy-remove-kubeadmin.yaml
+++ b/community/SC-System-and-Communications-Protection/policy-remove-kubeadmin.yaml
@@ -18,15 +18,11 @@ spec:
         spec:
           remediationAction: inform
           severity: low
-          namespaceSelector:
-            exclude:
-              - kube-*
-            include:
-              - default
           object-templates:
             - complianceType: mustnothave
               objectDefinition:
                 kind: Secret
+                apiVersion: v1
                 metadata:
                   name: kubeadmin
                   namespace: kube-system


community/SC-System-and-Communications-Protection/policy-remove-kubeadmin.yaml
```diff
@@ -19,17 +19,15 @@ spec:
           remediationAction: inform
           severity: low
           namespaceSelector:
-            exclude:
-              - kube-*
             include:
-              - default
+              - kube-system
```
dhaiducek: For clarity, I think we would prefer to omit the `namespaceSelector` entirely here in favor of defining it in the Secret directly. (When the namespace is provided in the `objectDefinition`, it overrides the `namespaceSelector`.)

dhaiducek (changes requested):
Thanks for this update! I agree this `namespaceSelector` doesn't make sense here. I have a comment about defining it more clearly--let me know what you think!

dhaiducek:
In case you're new to `DCO`, you can add a signoff to the most recent commit using the CLI:
```
git commit --signoff --amend
git push --force
```

dhaiducek (approved):
LGTM!

PR #327, opened 2022-08-11
Address problem downloading the cli image in the OPP policyset

gparvin:
The openshift CLI image sometimes can't be downloaded.  In one scenario
I recreated the issue and noticed one pod successfully downloaded the
image and another pod didn't.  Both pods were on the same node so I
deleted the pod that failed to download and when the pod restarted it
worked find.  My guess is there may be a permissions or timing issue
that caused the problem so I have re-ordered some resources to change
the timing.

Currently this is a guess so it may turn into some job configuration.

### DIFF
--- a/policygenerator/policy-sets/community/openshift-plus/input-quay/policy-config-quay.yaml
+++ b/policygenerator/policy-sets/community/openshift-plus/input-quay/policy-config-quay.yaml
@@ -5,6 +5,50 @@ metadata:
     openshift.io/cluster-monitoring: "true"
   name: local-quay
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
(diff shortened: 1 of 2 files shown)


ascerra:
@gparvin I've tested this PR against deploying OPP on OCP 4.10.26 on AWS using m4.xlarge as my instance type and we did not hit the same image pull authentication issue as before. Prior to this PR this same deployment configuration led to that issue.

I'll be running additional deployments and will update again here once I have more data but so far so good.

Update: This was tested on OCP 4.10.22 as well and it worked as expected.

gparvin:
Thanks for the feedback
/unhold

mprahl (approved):

PR #329, opened 2022-08-21
Namespace selector updates

dhaiducek:
- For clusterwide objects, remove it entirely
- For objects with namespace specified, remove it entirely
- For selectors that have a definitive `include`, remove `exclude`

### DIFF
--- a/community/AC-Access-Control/policy-roles-no-wildcards.yaml
+++ b/community/AC-Access-Control/policy-roles-no-wildcards.yaml
@@ -19,7 +19,6 @@ spec:
           remediationAction: inform # will be overridden by remediationAction in parent policy
           severity: high
           namespaceSelector:
-            exclude: ["kube-*"]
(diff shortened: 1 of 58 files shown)


gparvin (approved):
Thanks for making this improvement to help clarify where these selectors aren't needed -- which is a lot of our policies.  I wonder if we need more good namespace selector samples -- especially around NetworkPolicies and maybe Routes,etc...  certs are generally good samples too

dhaiducek:
Yes--having namespaceSelector-specific examples would be nice. I have a couple thoughts:
- Maybe a namespaceSelector doc page? I have a doc PR to bolster its description, but I'm not sure if we want a dedicated page on it.
- I should probably compose a small blog about the selector (possibly in lieu of the doc page). It could even reference the Gatekeeper policy with the labeled namespace webhook and build on it to use the label key in a policy.

PR #330, opened 2022-08-22
Moving the secrets and securedcluster cr before job broke acs install

gparvin:
The ACS job has to be defined before the secrets and secured cluster
CR since object processing does not continue once a failure is
encountered.  This could be something to think about enhancing in the
config policy controller -- but for now I'm just undoing the change to
this file.  The change to ODF is working and doesn't have this issue.

### DIFF
--- a/policygenerator/policy-sets/community/openshift-plus/input-sensor/policy-acs-central-ca-bundle.yaml
+++ b/policygenerator/policy-sets/community/openshift-plus/input-sensor/policy-acs-central-ca-bundle.yaml
@@ -63,62 +63,6 @@ subjects:
   name: create-cluster-init
   namespace: stackrox
 ---
-apiVersion: platform.stackrox.io/v1alpha1
-kind: SecuredCluster
-metadata:
-  name: stackrox-secured-cluster-services
-  namespace: stackrox
-spec:
-  clusterName: |
-    {{ fromSecret "open-cluster-management-agent" "hub-kubeconfig-secret" "cluster-name" | base64dec }}
-  auditLogs:
-    collection: Auto
-  centralEndpoint: |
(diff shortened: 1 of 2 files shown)


JustinKuli:
> object processing does not continue once a failure is encountered

Do you mean, later object-templates in the generated ConfigurationPolicy were not being handled because the SecuredCluster CRD was not yet installed?

Which version did you notice this in? I would expect https://github.com/open-cluster-management-io/governance-policy-template-sync/pull/123/commits/212856173d09382a1bd104afbb12e43daa2799f7 to resolve the issue I described above.

JustinKuli:
Sorry, my mistake. The commit I linked earlier is for what sounded like a similar problem, but with template-sync, not config-policy. And actually, I think the problem here is in template rendering (the `{{ fromSecret ... }}` stuff)?

If so, did https://github.com/open-cluster-management-io/config-policy-controller/pull/47/files actually make this worse? After that commit, I think all templates in all the object-templates are handled at once, and it looks like any errors would return early (bad)... previously each object template did its templating just before being checked/applied.

gparvin:
/hold
@JustinKuli You are correct -- now that I'm testing this on ACM 2.6 I see it behaves worse.  I'll see if switching to different policy templates helps.

gparvin:
I added a change to set `consolidateManifests` to false which I think is the fix I have to go with.  It seemed to work fine but I'm not sure if there will still be a timing issue.

mprahl (approved):

mprahl:
@gparvin an alternative would be to use `lookup` instead of `fromSecret` since `lookup` does not raise an error when the secret doesn't exist.

This is fine though.

PR #331, opened 2022-08-23
Remove subscriptions that ODF now manages on its own

gparvin:
ODF should install correctly now without specifying the OCS or MCG
subscriptions.


gparvin:
ODF came up fine with this change.
/unhold

JustinKuli (approved):

PR #332, opened 2022-09-02
Need separated config policies when using acm 2.6 with stackrox

gparvin:
Changes in ACM 2.6 prevent object definitions from loading properly.
This is similar to a recent case where the problem happened with
stackrox on the hub.  This is the managed cluster case which only
happens when you have a managed cluster separate from the hub.


dhaiducek (approved):
LGTM!

PR #336, opened 2022-09-12
Update the kyverno configmap to ignore control plane namespaces

gparvin:
I'm including OpenShift Platform Plus components as control plane namespaces to ignore. OPP namespaces as configured by the OPP policyset

Refs:
 - https://github.com/stolostron/backlog/issues/25674


gparvin:
/hold
Christian requested adding argocd namespace so need to add that too

JustinKuli (commented):
Seems good, although it's a hassle to read and it looks like you snuck in some things other than namespaces ;)

Can you add a reason for the `webhooks` change to the commit message?

Also, separate question, there are a few filters like `[*,open-cluster-*,*]`, which seem like they *should* catch both `open-cluster-management` and `open-cluster-management-agent` namespaces. But I couldn't immediately find any docs on what kind of expressions can go there - do you have any pointers? Or if you've verified it works, that should be fine. I'm just weary of these kinds of wildcards when they're not fully documented and I don't know the implementation (compare to the recent-ish changes in policy namespaceSelector).

gparvin:
> Seems good, although it's a hassle to read and it looks like you snuck in some things other than namespaces ;)
> 
> Can you add a reason for the `webhooks` change to the commit message?
> 
> Also, separate question, there are a few filters like `[*,open-cluster-*,*]`, which seem like they _should_ catch both `open-cluster-management` and `open-cluster-management-agent` namespaces. But I couldn't immediately find any docs on what kind of expressions can go there - do you have any pointers? Or if you've verified it works, that should be fine. I'm just weary of these kinds of wildcards when they're not fully documented and I don't know the implementation (compare to the recent-ish changes in policy namespaceSelector).

Good points:
1. Webhooks change: I installed kyverno and looked at what we had in our policy compared to what's in the configmap by default in kyverno 1.7.3.  The webhooks line is a straight copy of what kyverno https://github.com/kyverno/kyverno/blob/main/charts/kyverno/templates/configmap.yaml#L25
2. Filter docs are not very helpful: https://kyverno.io/docs/installation/#configmap-flags  - I did verify the wildcards are working

JustinKuli:
From what I can see, the filters likely use https://github.com/kyverno/go-wildcard. And it looks like that works with `*` in the middle of the filter. (Unlike what we previously had for our wildcards in namespaceSelector, which only worked at the beginning or end.)

JustinKuli (approved):

ch-stark:
we also need stackrox namespace

gparvin:
@ch-stark The stackrox namespace is in the filter list.  Also, argocd uses openshift-gitops namespaces so it will be filtered automatically since it starts with openshift.

I'm not going to merge this since this branch is being used for a demo tomorrow.

gparvin:
It's my other PR that has been forked to use in tomorrow's demo.  This one can be delivered now so unholding
/unhold

PR #343, opened 2022-09-22
Update README for Openshift Plus community policyset

ascerra:
I've included that this is specific to AWS for now. Added some additional info that I was missing when I first attempted to deploy (subscription admin privs, and managedclustersetbinding).

Signed-off-by: Adam Scerra <ascerra@redhat.com>


gparvin (approved):
Thanks!

PR #346, opened 2022-09-23
Some kyverno policies should not do background checking

gparvin:
Any kyverno policy that is working with cluster scoped resources will potentially match lots of resources if `background: true` is set in the ClusterPolicy.  Example: any of the policies that look at ClusterRoles. If the resource is namespaced we do some namespace filtering to keep from matching the infrastructure resources -- but we can't do that filtering for all cluster scoped resources (maybe we could do some -- but prefer to just turn background off)

Refs:
 - https://github.com/stolostron/backlog/issues/25864


gparvin:
Let me know if any of these seem like background should be `true`

ChunxiAlexLuo (approved):
/lgtm

/hold

PR #349, opened 2022-09-28
Links and timing concerns for quay

gparvin:
This commit adds the links for:
- acm observability grafana
- ACS
- quay

This commit re-arranges the quay and ACS resources.  The ACS resources that contain templates are now separated into a sync resources policy which will need to always have `consolidateManifests` set to false. The quay hub install now has an install and a config policy.  The config policy has some templates in the first 2 resources and the `consolidateManifests` is not set (so it's true) and this should solve the quay timing issue with the pull secret if using ACM 2.6 or newer.

Refs:
 - https://github.com/stolostron/backlog/issues/25659
 - https://github.com/stolostron/backlog/issues/25923

### DIFF
--- a/policygenerator/policy-sets/community/openshift-plus/input-acm-observability/policy-ocm-console.yaml
+++ b/policygenerator/policy-sets/community/openshift-plus/input-acm-observability/policy-ocm-console.yaml
@@ -1,0 +1,11 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: observability
+spec:
+  applicationMenu:
+    section: Red Hat applications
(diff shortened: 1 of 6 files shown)


gparvin:
/hold
Testing this some more.

gparvin:
/unhold
I think these changes are good.

policygenerator/policy-sets/community/openshift-plus/input-acm-observability/policy-ocm-console.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: console.openshift.io/v1
```
ChunxiAlexLuo, line 1: https://docs.openshift.com/container-platform/4.6/web_console/customizing-the-web-console.html#creating-custom-links_customizing-web-console

ChunxiAlexLuo (approved):
/lgtm

PR #352, opened 2022-10-07
[TVK Policies] Added backup retention and scheduling policy creation, changed min Openshift version to 4.8

sachin-trilio:
Added backup retention and scheduling policy creation
Changed min Openshift version to 4.8 in README as per review comments


sachin-trilio:
Requesting review @gparvin

gparvin (approved):
Thanks!

PR #353, opened 2022-10-14
Update to help anyone that wants to use the quay bridge operator

gparvin:
This is a mostly untested template of how the quay bridge operator could be deployed.  I'm not sure if it needs to be on the hub cluster but it's currently commented out with some instructions due to a manual procedure needed for creating a token.

Refs:
 - https://github.com/stolostron/backlog/issues/25954


dhaiducek (approved):
LGTM! The enhancement to get objects from other namespaces would really simplify things here!

PR #360, opened 2022-10-27
Adding policies for Node scaling operations

ttracey-rh:
Signed-off-by: Tom Tracey <ttracey@redhat.com>

Adding two policies to create new BMH and NMState configs for an existing ZTP managed cluster in ACM.


gparvin (approved):
lgtm

PR #361, opened 2022-11-02
Duplication cleanup

JustinKuli:
A few files were re-using names, and one policy had the `severity` set at the wrong level, sometimes causing an error.


gparvin (approved):
lgtm

PR #362, opened 2022-11-03
Added a second multitenancy policyset for hub only policies

gparvin:
Some of the multitenancy policies are for the hub and some are for all managed clusters.  I tried to separate them. Be aware I did not modify the placement settings which currently have all policies deployed only to the hub, but this change will help anyone looking at this understand what they need to do.

Refs:
 - https://github.com/stolostron/backlog/issues/25956


gparvin:
I think this will need to be rebased after the community branch merges.  I'll just close this for now.

PR #368, opened 2022-11-17
spec: and its subheadings in the Policy is wrong

chuckersjp:
The `spec:` heading and it sub items are not part of `metadata:` item so this policy doesn't perform as expected.

The `spec:` section of the Policy should be shifted left 2 spaces to line up under `metadata:`

### DIFF
--- a/community/CM-Configuration-Management/policy-ingress-controller.yaml
+++ b/community/CM-Configuration-Management/policy-ingress-controller.yaml
@@ -26,11 +26,11 @@ spec:
                 metadata:
                   name: default
                   namespace: openshift-ingress-operator
-                  spec:
-                    nodePlacement:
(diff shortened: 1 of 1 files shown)


dhaiducek:
@chuckersjp Thanks for the update! Would you please squash the commits and add a DCO signoff to the commit? This project uses DCO in lieu of a contributor license agreement.

These both can be accomplished using `git rebase -i --signoff` and specifying "squash" in the list of commits that appears. Let me know if you need any further guidance on it.

chuckersjp (commented):
Signed-off-by: Chuck Douglas <cdouglas@redhat.com>

chuckersjp:
Gonna try doing this from scratch.

PR #369, opened 2022-11-17
`spec:` line and subs are wrong

chuckersjp:
The `spec:` line and subheadings should be even with `metadata:` as they are an item of their own.

Without this change, the policy doesn't actually apply anything which is not the intended effect.


dhaiducek (approved):
Thanks for the update, @chuckersjp!

PR #370, opened 2022-11-17
Add tolerations to allow pods to start

chuckersjp:
This change adds tolerations to the policy to allow the router pods to spin up on infra labelled nodes.


ChunxiAlexLuo (approved):
Here without `key`, tolerations will ignore any taint.

Add @gparvin additional review

/hold

gparvin (approved):

PR #371, opened 2022-11-18
Prep for move to community repository

gparvin:
This creates some simple API validation (which could be enhanced more) and changes ACM specific usage to open cluster management.

I have not changed references to stolostron yet.  I'm thinking that will be a PR to do after the move to avoid breaking links now. Any feedback or PR updates are welcome.

Refs:
 - https://github.com/stolostron/backlog/issues/25068


dhaiducek:
I'm not so sure we should make these changes here midstream since we may want to preserve Red Hat references here in `stolostron`. It seems we should migrate the repo, fork it back, and then make the changes upstream to reflect it being upstream?

dhaiducek:
The changes for upstream overall look good, and that schema validator is a really nice addition! I might move it to a more generic "builds" or "scripts" folder so that we can put other scripts there in the future, though.

policygenerator/policy-sets/stable/README.md
```diff
@@ -8,4 +8,4 @@ Kustomize plugin for more information on using the policy generator.
 
 Policy  | Description | Prerequisites
 ------- | ----------- | -------------
-[Advanced Cluster Management Hardening](./openshift-hardening) | Applies best practices for how to harden your Advanced Cluster Management hub installation. | Needs to be deployed only to the Advanced Cluster Management hub cluster. The `PolicySet` uses cluster `Placement` and not the `PlacementRule` placement mechanism.
+[Open Cluster Management Hardening](./openshift-hardening) | Applies best practice policies for your Open Cluster Management hub installation. | Needs to be deployed only to the Open Cluster Management hub cluster. The `PolicySet` uses cluster `Placement` and not the `PlacementRule` placement mechanism.
```
dhaiducek:
```suggestion
[Open Cluster Management Hardening](./ocm-hardening) | Applies best practice policies for your Open Cluster Management hub installation. | Needs to be deployed only to the Open Cluster Management hub cluster. The `PolicySet` uses cluster `Placement` and not the `PlacementRule` placement mechanism.
```

dhaiducek: Along with this, we'd need to update the name of the directory.

dhaiducek (changes requested):

policygenerator/policy-sets/stable/README.md
gparvin: @dhaiducek I made changes related to renaming the hardening directory for the community policy and fixed the bad link.  I can't rename stable because we support stable and need to keep the same name until we create a `release-2.6` branch with the supported policy there.  Then we can do the rename and update the automation that tests the stable policy.

gparvin: @dhaiducek As discussed I have removed the OCM specific changes for now with the goal of using magic mirror so we can have differences in upstream and midstream readme content.  Take a look to see if things look ok now.
Thanks!

build/validate-policies.sh
```diff
@@ -0,0 +1,75 @@
...
+
+if [ $RC -ne 0 ]; then
+	go install github.com/yannh/kubeconform/cmd/kubeconform@$KC_VERSION
+fi
+
+if [ ! -d schemas ]; then
+	echo "Getting Schema conversion tool"
+	mkdir schemas
+	cd schemas
+	curl -o crd-schema.py https://raw.githubusercontent.com/yannh/kubeconform/master/scripts/openapi2jsonschema.py
```
mprahl: Could you please use a tagged version or a specific commit?

build/validate-policies.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "Checking stable policies"
+validatePolicies stable
+
+echo "Checking community policies"
+validatePolicies community
+
+if [ ! -f kustomize ]; then
+	echo "Downloading kustomize"
+	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
```
mprahl: Could you use `go install` instead? Also please use a tagged version.

build/validate-policies.sh
```diff
@@ -0,0 +1,75 @@
...
+fi
+
+if [ ! -f ~/.config/kustomize/plugin/policy.open-cluster-management.io/v1/policygenerator/PolicyGenerator ]; then
```
mprahl: Could you set the `KUSTOMIZE_PLUGIN_HOME` environment variable instead of using the home directory to store the Policy Generator?

gparvin: ok, I got this working with the latest changes.  Thanks for suggesting this.

build/validate-policies.sh
gparvin: Done

build/validate-policies.sh
gparvin: Done

build/validate-policies.sh
```diff
@@ -0,0 +1,77 @@
...
+
+echo "Checking stable policies"
+validatePolicies stable
+
+echo "Checking community policies"
+validatePolicies community
+
+if [ ! -f kustomize ]; then
+	echo "Installing kustomize"
+	GOBIN=$(pwd)/ GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v4@$KUSTOMIZE_VERSION
```
dhaiducek: For other flows, we use:
```
export GOBIN=${PWD}/bin
export PATH := ${GOBIN}:${PATH}
```

Maybe use that at the top of the file instead? (Especially since eventually we may want to do a `go install`  here once the generator is migrated upstream.)

build/validate-policies.sh
dhaiducek: Given the change, shouldn't this line be:
```suggestion
if [ ! -f ${KUSTOMIZE_PLUGIN_HOME}/policy.open-cluster-management.io/v1/policygenerator/PolicyGenerator ]; then
```

mprahl:
@gparvin could you please squash your commits?

build/validate-policies.sh
mprahl: @gparvin you'll need to update this if statement now to use the `KUSTOMIZE_PLUGIN_HOME` environment variable value.

dhaiducek (commented):
Nit: I'm still not convinced the ACM references should be updated to OCM in this PR?

build/validate-policies.sh
gparvin: Thanks, I've made updates to use GOBIN.

gparvin:
> Nit: I'm still not convinced the ACM references should be updated to OCM in this PR?

@dhaiducek I thought I had removed those references.  As you pointed out there were some more which I have now removed.  Thanks!

build/validate-policies.sh
```diff
@@ -0,0 +1,82 @@
...
+	echo "Getting Schema conversion tool"
+	mkdir schemas
+	cd schemas
+	curl -o crd-schema.py https://raw.githubusercontent.com/yannh/$KUBECONFORM/$KC_VERSION/scripts/openapi2jsonschema.py
+	chmod a+x crd-schema.py
+
+	echo "Converting CRDs to Schemas"
+	./crd-schema.py https://raw.githubusercontent.com/open-cluster-management-io/governance-policy-propagator/main/deploy/crds/policy.open-cluster-management.io_policies.yaml
+	./crd-schema.py https://raw.githubusercontent.com/open-cluster-management-io/governance-policy-propagator/main/deploy/crds/policy.open-cluster-management.io_placementbindings.yaml
+	./crd-schema.py https://raw.githubusercontent.com/open-cluster-management-io/api/main/cluster/v1beta1/0000_02_clusters.open-cluster-management.io_placements.crd.yaml
```
dhaiducek: These should probably point to `stolostron` also. Maybe put it in a variable here and then update the workflow with `export GITHUB_ORG=${{ github.repository_owner }}`? That way, all files can be the same in all orgs.
```suggestion
	GITHUB_ORG=${GITHUB_ORG:-"open-cluster-management-io"}
	./crd-schema.py https://raw.githubusercontent.com/${GITHUB_ORG}/governance-policy-propagator/main/deploy/crds/policy.open-cluster-management.io_policies.yaml
	./crd-schema.py https://raw.githubusercontent.com/${GITHUB_ORG}/governance-policy-propagator/main/deploy/crds/policy.open-cluster-management.io_placementbindings.yaml
	./crd-schema.py https://raw.githubusercontent.com/${GITHUB_ORG}/api/main/cluster/v1beta1/0000_02_clusters.open-cluster-management.io_placements.crd.yaml
```

build/validate-policies.sh
```diff
@@ -0,0 +1,82 @@
...
+if [ ! -f ${GOBIN}/kustomize ]; then
+	echo "Installing kustomize"
+	GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v4@$KUSTOMIZE_VERSION
+fi
+
+export KUSTOMIZE_PLUGIN_HOME=${GOBIN}
+if [ ! -f ${KUSTOMIZE_PLUGIN_HOME}/${GENERATOR_PATH}/PolicyGenerator ]; then
+	echo "Downloading the generator"
+	PLATFORM=`uname | tr '[:upper:]' '[:lower:]'`
+	curl -L -o generator https://github.com/stolostron/policy-generator-plugin/releases/latest/download/${PLATFORM}-amd64-PolicyGenerator
```
dhaiducek, line 87: Side note: We seriously need to get the generator upstream...

dhaiducek (commented):
Looking good! I don't recall if I said it before, but thanks for all the link updates!

I have a suggestion around dynamically setting the GitHub org.

build/validate-policies.sh
gparvin: I made an update to use a common org.  Hopefully it will match the right one if I understand the doc correctly. THanks!

dhaiducek (changes requested):
I didn't think about the environment variable already being available! Does removing the new `export` and using `GITHUB_REPOSITORY_OWNER` instead of `GITHUB_ORG` work?

gparvin:
> I didn't think about the environment variable already being available! Does removing the new `export` and using `GITHUB_REPOSITORY_OWNER` instead of `GITHUB_ORG` work?

@dhaiducek I changed it as requested but take a look and see if it's ok now.

dhaiducek (approved):
LGTM! Thanks, @gparvin!

PR #373, opened 2022-11-30
Add Placement documentation

dhaiducek:
I'm always confused when Placement comes into play. Add some links to clarify (partially for my own benefit!).

Also updates all old links to use `2.6` docs.


gparvin (approved):
Thanks so much for making these fixes!

PR #379, opened 2022-12-02
Added OADP

tkagn:
Signed-off-by: Stephen Wilson <swilson@redhat.com>


community/CM-Configuration-Management/policy-oadp.yaml
```diff
@@ -0,0 +1,142 @@
...
+            - complianceType: musthave
+              objectDefinition:
+                kind: Secret
+                apiVersion: v1
+                metadata:
+                  name: cloud-credentials
+                  namespace: openshift-adp
+                data:
+                  cloud: >-
+                    W2RlZmF1bHRdCmF3c19hY2Nlc3Nfa2V5X2lkPQphd3Nfc2VjcmV0X2FjY2Vzc19rZXk9CgoK
```
JustinKuli: Should there be a comment explaining what this should contain? Right now it decodes to
```
[default]
aws_access_key_id=
aws_secret_access_key=
```
I assume that would need to be filled in by the user?

tkagn: Correct. That is just a placeholder/template to be filled out by the administrator.

community/CM-Configuration-Management/policy-oadp.yaml
```diff
@@ -0,0 +1,142 @@
...
+                spec:
+                  targetNamespaces:
+                    - openshift-adp
+          remediationAction: inform
+          severity: high
+    - objectDefinition:
+        apiVersion: policy.open-cluster-management.io/v1
+        kind: ConfigurationPolicy
+        metadata:
+          name: policy-odf-operator-subscription
```
gparvin:
```suggestion
          name: policy-oadp-operator-subscription
```

community/CM-Configuration-Management/policy-oadp.yaml
gparvin: @tkagn We can't recommend a policy that directly includes secret content in the policy.  This is because any etcd backups would only encrypt secret resources -- not the policy resources.  Additionally all ACM must-gathers will collect policy content further exposing the secret values.  The recommended way to include secret content in policies is to follow the guidance in the **Important** note here: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6/html/governance/governance#hub-templates

Along with a new policy make sure to include an entry in the README.md file too.  Thanks!

tkagn: @gparvin Updated the policy and added entry to README.md

community/CM-Configuration-Management/policy-oadp.yaml
tkagn: @gparvin Corrected. Thanks.

gparvin (approved):

PR #383, opened 2022-12-19
Add disable self-provisioner policy

chuckersjp:

This change adds a policy that will disable the project self-provisioner ability from a cluster.

It is currently set to `inform` rather than `enforce` as a default.


ch-stark:
thanks a lot!

gparvin (commented):
Can you add an entry in the README.md file in the right section of the table that provides some explanations of what the policy does?  The readme is in the community folder.  Thanks!

chuckersjp:
Added update to README.md

gparvin (approved):

PR #384, opened 2022-12-22
Create a policy to install and configure OpenShift Service Mesh

fperearodriguez:
Add a new policy to install and configure a basic Service Mesh in OpenShift Container Platform.

This policy installs the required Operators and configures a basic Service Mesh.


gparvin (approved):
Thanks!  This looks nice!

PR #391, opened 2023-01-24
Fix trilio validation error and update aap policy

gparvin:
The validation script is failing on a new policy that didn't run the validation because the PR was created before validation was added.  I also removed the channel in the AAP policy so the latest will be used.


dhaiducek (approved):
LGTM! Thanks, Gus!

PR #396, opened 2023-02-13
Configure GitOps with the policy generator

dhaiducek:


community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
```diff
@@ -0,0 +1,138 @@
...
+    apiGroup: policy.open-cluster-management.io
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
+metadata:
+  name: placement-policy-openshift-gitops-policygenerator
+spec:
+  clusterConditions:
+  - status: "True"
+    type: ManagedClusterConditionAvailable
```
JustinKuli: Do we still want/need this cluster condition? I'm assuming it was copied from other policies, but it caused problems with `pruneObjectBehavior`, so I think we stopped using it in default placements... I know that pruning isn't desired in this policy, but if this condition isn't necessary, I'd rather not have it.

JustinKuli (commented):
This is really cool!

community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
dhaiducek: Definitely copy/paste. I'll take it out.

community/CM-Configuration-Management/policy-openshift-gitops.yaml
```diff
@@ -21,12 +21,11 @@ spec:
           object-templates:
             - complianceType: musthave
               objectDefinition:
-                # This is an auto-generated file. DO NOT EDIT
                 apiVersion: operators.coreos.com/v1alpha1
                 kind: Subscription
                 metadata:
                   name: openshift-gitops-operator
-                  namespace: openshift-operators
+                  namespace: openshift-gitops-operator
```
JustinKuli: Sorry I missed this the first time, where did this change come from?

dhaiducek: Hmmm. I'll revert it. For some reason our cluster today had it in this namespace...

community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
```diff
@@ -0,0 +1,135 @@
...
+    apiGroup: policy.open-cluster-management.io
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
+metadata:
+  name: placement-openshift-gitops-policygenerator
+spec:
+  clusterSelector:
+    matchExpressions:
+      - {key: environment, operator: In, values: ["dev"]}
```
gparvin: Since this only applies to the hub, can we provide that specific placement instead of our normal default?
```suggestion
      - {key: name, operator: In, values: ["local-cluster"]}
```

dhaiducek: Good point!

JustinKuli (approved):

PR #399, opened 2023-03-20
Create a stable opp policyset and the setup policyset for opp

gparvin:
First pass at an initial stable policyset for OPP.  This requires ACM 2.7 because it uses the new dependencies feature.  I'm also creating a community policyset that will install prereqs for OPP on AWS


gparvin:
/hold
Still working on testing this.

gparvin:
/unhold
FYI: @ascerra This contribution of the setup for OPP policy set may be "closer" to something for general use instead of the demo policy I pointed you to.  Note that this uses 3 availability zones in AWS by default instead of 1 and is using 1 `m6a.4xlarge` compute in each zone.

gparvin:
/hold
Adding some extra configuration helpers for AWS

gparvin:
/unhold
Tested on us-east-1 and 2.  The next simplification that would help is to use a range over the zones so their customization can be different per region.  Right now the us-east-1c zone seems to have lots of capacity issues.  Better customization might help avoid this.

gparvin:
@ascerra Note that this PR creates the setup PolicySet for OPP which can create the MachineSet for ODF in us-east-1 or us-east-2 with the appropriate AWS ami-id automatically.  Right now the zones are limited to zone a, b and c in both regions -- the problem I have is sometimes I can't resume the cluster because of capacity issues in us-east-1 zone c but as long as you're not hibernating I think you will be ok with this config.

policygenerator/policy-sets/stable/openshift-plus/scripts/acs-bundle-secret.sh
```diff
@@ -0,0 +1,145 @@
...
+            echo "Invalid input: ${1}" >&2
+            exit 1
+            shift
+            ;;
+        esac
+done
+set -- "${POSITIONAL[@]}" # restore positional parameters
+
+ACS_HOST="$(oc get route -n $CENTRALNS central -o custom-columns=HOST:.spec.host --no-headers):443"
+if [[ -z "$ACS_HOST" ]]; then
```
JustinKuli: This will always be true, because it ends with `:443` above, even if the route wasn't found.

JustinKuli (commented):
One small thing I noticed (among a big PR). Dependencies and policies looked right from what I saw.

policygenerator/policy-sets/stable/openshift-plus/scripts/acs-bundle-secret.sh
gparvin: Sorry about that, the scripts directory was a backup of the original scripts I had used which aren't needed anymore.  I deleted them.  Thanks for taking a look!

JustinKuli (approved):

gparvin:
This addressed https://issues.redhat.com/browse/ACM-1278

PR #400, opened 2023-03-23
Create policies to setup sso across managedclusters

ckandag:
Signed-off-by: Chaitanya Kandagatla <ckandaga@redhat.com>

related to :
https://issues.redhat.com/browse/ACM-4188


JustinKuli (approved):
/hold for other reviews

This looks good to me! Pretty cool to see what can be done with the raw templates.

I think I see what you were talking about with the policy namespace being required in these, so that the dependencies can work correctly. What you have here looks good to me in that respect.

If cross-namespace secret lookup is possible eventually, then I think all of these policies could use the same namespace? And then maybe it would work as a PolicySet? But for now, I think this is right.

gparvin (approved):

PR #405, opened 2023-04-04
Update comments and warning

chuckersjp:
This change adds a warning in the comments about ensuring that the policy name and namespace must be less than 63 characters in length.

It also adds `namespace: <POLICIESNS>` to the policies and the `PlacementRules` and `PlacementBinding`


community/CM-Configuration-Management/policy-autoscaler-templatized.yaml
```diff
@@ -12,12 +12,16 @@
...
+#
+# Also note that the sum of characters of the Policy name and the <POLICIESNS> must be less than 63 characters or it will fail
+# validation.
 
 
 apiVersion: policy.open-cluster-management.io/v1
 kind: Policy
 metadata:
   name: policy-autoscaler-templatized
+  namespace: <POLICIESNS>
```
gparvin, line 24: We don't want a default namespace to be provided with policies.  Leaving the namespace unspecified is desirable so the user can deploy it to the namespace they choose.

chuckersjp: Shouldn't all policies go into a define namespace though?  Or is it just an issue with this particular policy?

gparvin: If applying the policy using the CLI the user will specify the namespace using `oc apply -f policy.yaml -n namespace`.  If the user deploys using gitops the namespace is provided as part of kustomize settings generally so leaving it out of the policy is our standard best practice.  Thanks!

chuckersjp: Okay.  Fair enough.  I have currently having some problems with it not appearing where it needs to go.  I will look at it some more.

dhaiducek:
A webhook as well as UI validation are in place in more recent versions to prevent lengthy policy name/namespace combinations from being created. Given that, could we consider closing this PR?

chuckersjp:
Sure.  If this is covered elsewhere I am good with closing it.

PR #409, opened 2023-04-21
create policies to backup and restore persistent apps

birsanv:
The Policies available here provide backup and restore support for stateful applications running on managed clusters or hub. Velero is used to backup and restore applications data. 
Policies:
- oadp-hdr-app-install installs oadp and configure connection to the storage on managed clusters
- oadp-hdr-app-backup backups applications on managed clusters
- oadp-hdr-app-restore applications on managed clusters

The configmap  hdr-app-configmap in the PR sets configuration options for the backup storage location, for the backup schedule backing up applications, and for the restore resource used to restore applications backups.

https://issues.redhat.com/browse/ACM-4708

PR #410, opened 2023-05-04
Add missing CRDs to validate script

gparvin:
We haven't needed `PolicySet` or `PolicyAutomation` in the PR validation until now.  We should allow these CRs to be candidates for community policy submissions so I have added them to the script.

Refs:
 - https://github.com/open-cluster-management-io/policy-collection/pull/409


dhaiducek (changes requested):
Since we're here...would you also update the policy generator link to use `GITHUB_REPOSITORY_OWNER` now that we have forked versions?

gparvin:
@dhaiducek Sure thing!   Updated the link.

dhaiducek (approved):
/hold Thanks! I realized now that I see the update that `stolostron` doesn't have GitHub releases, so we can either roll with this and figure that out later or set it directly to `open-cluster-management-io` for now...

gparvin:
The command worked for me so will merge this for now.
/unhold

PR #411, opened 2023-05-04
Removing policy dependencies

levenhagen:
I think having this dependency in this policy won't be ok because the other previous SSO policies are not deployed to managed clusters, so the dependency will stop this policy from being executed in managed clusters other than the local-cluster.


dhaiducek:
@levenhagen Sorry for the delay here! Would you please add the DCO signoff to this PR? Since it's a single commit, doing an "amend" and then a force push should do it:
```
git commit --amend --signoff
git push --force
```

PR #415, opened 2023-05-17
Stateful app backup policies

sahare:
The Policies available here provide backup and restore support for stateful applications running on managed clusters or hub. Velero is used to backup and restore applications data.
Policies:

oadp-hdr-app-install installs oadp and configure connection to the storage on managed clusters
oadp-hdr-app-backup backups applications on managed clusters
oadp-hdr-app-restore applications on managed clusters
The configmap hdr-app-configmap in the PR sets configuration options for the backup storage location, for the backup schedule backing up applications, and for the restore resource used to restore applications backups.

https://issues.redhat.com/browse/ACM-4708


birsanv:
this is a continuation of https://github.com/open-cluster-management-io/policy-collection/pull/409

dhaiducek (commented):
I'm discussing some things about this PR with our team before approving it. In particular, PlacementRule is deprecated, so while existing policies in the collection use PlacementRule, it might be good to go ahead and update this policy to use Placement now rather than having to revisit in the future. I'll confirm this, though.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policy-sets/acm-app-backup-policy-set.yaml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  name: acm-app-backup-placement
+spec:
+  clusterConditions: []
+  clusterSelector:
+    matchExpressions:
+      - key: acm-pv-dr
+        operator: In
+        values:
+          - backup
```
dhaiducek: Please update this to use Placement since PlacementRule is deprecated.

dhaiducek (changes requested):
Please update all PlacementRules (and PlacementBinding references in conjunction) to use Placement now that PlacementRule is deprecated.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-backup.yaml
```diff
@@ -0,0 +1,186 @@
...
+    policy.open-cluster-management.io/standards: NIST SP 800-53
+spec:
+  dependencies:
+  - apiVersion: policy.open-cluster-management.io/v1
+    compliance: Compliant
+    kind: Policy
+    name: oadp-hdr-app-install
+  disabled: false
+  policy-templates:
+    - objectDefinition:
```
mprahl, line 27: I just came in to say that there are a lot of `fromConfigMap` calls to the same `ConfigMap` and those API calls are not cached unfortunately. One alternative is to use `object-templates-raw` in each `ConfigurationPolicy` which is just a field that allows templating as long as the final output is valid YAML following the syntax of `object-templates`.

So you could use `with` from go text/template to store the value of the `ConfigMap` using the `lookup` template function. Then at least it'd only be one API call for the `ConfigMap` per `ConfigurationPolicy`.

mprahl: Note that object-templates-raw is only in ACM 2.7+.

birsanv: @mprahl can you add a sample please for object-templates-raw usage

mprahl: @birsanv not exactly your situation but here is an example:
```yaml
kind: Policy
metadata:
  name: label-configmaps
  namespace: policies
  annotations:
    policy.open-cluster-management.io/standards: NIST SP 800-53
    policy.open-cluster-management.io/categories: CM Configuration Management
    policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
spec:
  disabled: false
  policy-templates:
    - objectDefinition:
        kind: ConfigurationPolicy
        metadata:
          name: label-configmaps
        spec:
          remediationAction: enforce
          severity: low
          object-templates-raw: |
            {{- range (lookup "v1" "ConfigMap" "my-configmaps" "").items }}
            {{- if and .data.model (contains "Mustang" .data.model) }}
            - complianceType: musthave
              objectDefinition:
                kind: ConfigMap
                apiVersion: v1
                metadata:
                  name: {{ .metadata.name }}
                  namespace: {{ .metadata.namespace }}
                  labels:
                    ford.com/model: Mustang
            {{- end }}
            {{- end }}
```

birsanv: @mprahl this is quite a big change, is this needed to approve the PR ?

I could see a basic sample of the object-templates-raw with the documentation but I am not clear what changes need to be done to replace object-templates with object-templates-raw. 

>>So you could use with from go text/template to store the value of the ConfigMap using the lookup template function. Then at least it'd only be one API call for the ConfigMap per ConfigurationPolicy

birsanv: @mprahl I understand what the ask is, no need for a sample
Looking into making the requested changes, if they are needed for the PR to be approved..

birsanv: There are a few issues @mprahl when trying to use object-templates-raw + lookup instead of   object-templates + fromConfigMap

1. Since the user can create the policy on the hub under any namespace, the configmap used by the policy doesn't have a fixed namespace ( the configmap ns would be policy's ns ). So in the object-templates, the  fromConfigMap uses "" for the ns , for example '{{hub fromConfigMap "" "hdr-app-configmap" "backupNS" hub}}'.  This defaults to the policy namespace

If I change the  object-templates with  object-templates-raw, this is no longer working, I have to specify the namespace for the configmap

I have to change 
          ```
object-templates:
            - complianceType: mustnothave
              objectDefinition:
                apiVersion: velero.io/v1
                kind: Schedule
                metadata:
                  namespace: '{{hub fromConfigMap "" "hdr-app-configmap" "backupNS" hub}}'
                  labels:
                    cluster-id: '{{ fromClusterClaim "id.openshift.io" }}'
                    cluster-name: '{{ fromClusterClaim "name" }}'              
                status:
                  phase: FailedValidation                
          remediationAction: inform
          severity: high
```


with 

```
          object-templates-raw: |
            - complianceType: musthave
              objectDefinition:
                apiVersion: velero.io/v1
                kind: Schedule
                metadata:
                  namespace: '{{ (lookup "v1" "ConfigMap" "**app-policies**" "hdr-app-configmap").data.backupNS }}'
                  labels:
                    cluster-id: '{{ fromClusterClaim "id.openshift.io" }}'
                    cluster-name: '{{ fromClusterClaim "name" }}'
                status:
                  phase: FailedValidation                
          remediationAction: inform
          severity: high
```

2.  The configmap has keys in this format a.b ( for example backup.volumeSnapshotLocation ) which are supported by the fromConfigMap but not by the lookup since I assumes it takes this as a  path

Item 2 can be fixed, rename all the properties and take out the .
How do I implement point 1. though with the lookup function ?

Comment (mprahl) [reply community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-backup.yaml:27]: @birsanv if this is not an out of the box policy or commonly used policy, then the change is not required. If it will be commonly deployed, I want to minimize the impacts on the Kubernetes API.

For point 1, you can still use a mix of Hub templates and managed cluster side templates. That hasn't changed.

This is untested, but this is along the lines of what I was thinking of:
```yaml
      object-templates-raw: |
        {{hub with $configMap := (lookup "v1" "ConfigMap" "" "hdr-app-configmap") hub}}
        - complianceType: musthave
          objectDefinition:
            apiVersion: velero.io/v1
            kind: Schedule
            metadata:
              namespace: '{{hub $configMapdata.data.backupNS hub}}'
              labels:
                cluster-id: '{{ fromClusterClaim "id.openshift.io" }}'
                cluster-name: '{{ fromClusterClaim "name" }}'
            status:
              phase: FailedValidation     
          {{hub end hub}}           
      remediationAction: inform
      severity: high
```

Also, in ACM 2.8+, you can now use the `.ManagedClusterLabels` variable in Hub templates to potentially avoid `fromClusterClaim` calls on the managed cluster side.

For point 2, you can use the `index` template function to access properties/keys that have characters such as `.` and `-`.:
https://pkg.go.dev/text/template#hdr-Functions

birsanv: @mprahl thanks
I tried this template

```
    - objectDefinition:
        kind: ConfigurationPolicy
        metadata:
          name: check-schedule-failed-validation-2
        spec:
          object-templates-raw: |
            {{hub with $configMap := (lookup "v1" "ConfigMap" "" "hdr-app-configmap") hub}}
            - complianceType: musthave
              objectDefinition:
                apiVersion: velero.io/v1
                kind: Schedule
                metadata:
                  namespace: '{{hub $configMap.data.backupNS hub}}'
                  labels:
                    cluster-id: '{{ fromClusterClaim "id.openshift.io" }}'
                    cluster-name: '{{ fromClusterClaim "name" }}'
                status:
                  phase: FailedValidation     
              {{hub end hub}}                
          remediationAction: inform
          severity: high
```

and get the following error

    policy.open-cluster-management.io/hub-templates-error: >-
      failed to parse the template JSON string
      {"apiVersion":"policy.open-cluster-management.io/v1","kind":"ConfigurationPolicy","metadata":{"name":"check-schedule-failed-validation-2"},"spec":{"object-templates-raw":"{{hub
      with $configMap := (lookup \"v1\" \"ConfigMap\" \"\"
      \"hdr-app-configmap\") hub}}\n- complianceType: musthave\n 
      objectDefinition:\n    apiVersion: velero.io/v1\n    kind: Schedule\n   
      metadata:\n      namespace: '{{hub $configMap.data.backupNS hub}}'\n     
      labels:\n        cluster-id: '{{ fromClusterClaim \"id.openshift.io\"
      }}'\n        cluster-name: '{{ fromClusterClaim \"name\" }}'\n   
      status:\n      phase: FailedValidation     \n  {{hub end
      hub}}                \n","remediationAction":"inform","severity":"high"}}:
      template: tmpl:6: unexpected "\\" in operand

I don't understand the error message.. I tried to look for similar doc or samples using {{hub with .. }}  but can't find something similar to debug the issue

I am actually wondering if this will solve my issue at point 1 after all ( the lookup asks for the configMap namespace to be specified , a "" value is not looking at the policy namespace on the hub ; whereas the fromConfigMap with an "" ns gets the configmap from the policy namespace )
Wouldn't 
`{{hub with $configMap := (lookup "v1" "ConfigMap" "" "hdr-app-configmap") hub}} ` 

produce the same result as

`namespace: '{{ (lookup "v1" "ConfigMap" "" "hdr-app-configmap").data.backupNS }}'`

? 

The lookup above works only if I enter the ns name

`namespace: '{{ (lookup "v1" "ConfigMap" "app-policies" "hdr-app-configmap").data.backupNS }}'`

mprahl: To answer your question, `{{` is on the managed cluster side. So `app-policies` would only be there if the policy applied to a self-managed Hub. Hub templates get executed before the policy is copied to the managed cluster.

sahare:
@mprahl Please review, pushed some changes trying to address your request as follows:

- renamed properties that have .
- used the object-templates-raw instead of object-templates
- removed fromConfigMap, defined configMap and reused the cashed value in the templates

mprahl (dismissed):
Looks great! Thanks for making the changes.

/hold

mprahl:
> /retest

The failing CI is a GitHub workflow and not a Prow job, so `/retest` won't work for that.

sahare:
@mprahl We are done with testing, could you please re-approve and merge. Thanks!

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-backup.yaml
```diff
@@ -98,8 +98,7 @@ spec:
                     cluster-name: '{{ fromClusterClaim "name" }}'
                 status:
                   phase: Completed
-                  startTimestamp: '{{ (lookup "velero.io/v1" "Schedule" "{{hub $configMap.data.backupNS hub}}" "{{hub $configMap.data.backupPrefix hub}}-{{hub $configMap.data.backupVolumeSnapshotLocation hub}}-{{hub (printf "%s" .ManagedClusterName) hub}}").status.lastBackup }}'
-            {{hub end hub}}
+                  startTimestamp: '{{ (lookup "velero.io/v1" "Schedule" "{{hub $configMap.data.backupNS hub}}" "{{hub $configMap.data.backupPrefix hub}}-{{hub $configMap.data.backupVolumeSnapshotLocation hub}}{{hub end hub}}-{{hub (printf "%s" .ManagedClusterName) hub}}").status.lastBackup }}'
```
dhaiducek, line 101: Optional: It might be clearer if the `end` were outside of the `startTimestamp` value (unless there's a reason to place it here like maybe `.ManagedClusterName` no longer worked...) (this would apply to the remaining ConfigurationPolicies also)
```suggestion
                  startTimestamp: '{{ (lookup "velero.io/v1" "Schedule" "{{hub $configMap.data.backupNS hub}}" "{{hub $configMap.data.backupPrefix hub}}-{{hub $configMap.data.backupVolumeSnapshotLocation hub}}-{{hub (printf "%s" .ManagedClusterName) hub}}").status.lastBackup }}'
            {{hub end hub}}
```

dhaiducek (approved):
/hold in case you want to address the comment

Looks good! Thanks for this contribution, and thanks for your patience and iterations through the PlacementRule deprecation!

birsanv:
@dhaiducek the {{hub (printf "%s" .ManagedClusterName) hub}}") is outside the end block onpurpose, because it is not processed properly if it's inside .. :) you can verify, I get a nil value

Issue #407, opened 2023-04-20
ACS policies SecuredCluster


aakanmaz:
Hi 
it may be good to change collection method as  EBPF. KernelModule is deprecated 

https://docs.openshift.com/acs/3.74/release_notes/374-release-notes.html#kernel-collection-module

Added 20 March 2023

Currently, secured clusters can specify three options of collection methods for runtime events: eBPF (selected by default), kernel module, or no collection. Kernel module as a collection method is deprecated in the RHACS version 3.74 release and is planned for removal in the RHACS version 4.1 release.

dhaiducek:
@aakanmaz As the policy has been updated, would you please confirm whether this issue can be closed?

aakanmaz:
Hi 

It may be also good to check below files. Not sure but these seems still using KernelModule if I am not missing anything

‎policygenerator/policy-sets/community/openshift-plus/input-sensor/policy-acs-sync-resources.yaml
‎community/CM-Configuration-Management/policy-acs-operator-secured-clusters.yaml
‎policygenerator/policy-sets/stable/openshift-plus/output.yaml
‎policygenerator/policy-sets/community/openshift-plus/input-sensor/policy-advanced-managed-cluster-security.yaml

gparvin:
I addressed the additional concerns in pr #423 so closing now.

PR #423, opened 2023-08-01
Remove the community openshift-plus policy set which isn't being tested

gparvin:
Automation is testing the openshift-plus policy set under the stable folder.  I was going to keep the community policy set since it would work with older releases (pre-ACM 2.7).  Since it's not being tested we see it is quickly not going to be valuable so I'm removing it.

Refs:
 - https://github.com/open-cluster-management-io/policy-collection/issues/407


JustinKuli (approved):
Seems good to me! There are release branches in the stolostron fork, so users could always point to that if they need an unchanging version.

PR #431, opened 2023-10-06
fix bad format in audit config

ch-stark:
There was a bad intendation in configuring openshift-audit.
I verified the fix provided in my env

### DIFF
--- a/policygenerator/policy-sets/community/ocp-best-practices/input-audit/policy-config-audit.yaml
+++ b/policygenerator/policy-sets/community/ocp-best-practices/input-audit/policy-config-audit.yaml
@@ -9,4 +9,4 @@ spec:
       profile: WriteRequestBodies
     - group: system:authenticated
       profile: AllRequestBodies
-  profile: Default  
+    profile: Default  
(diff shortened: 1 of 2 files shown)


yiraeChristineKim (dismissed):

dhaiducek (changes requested):
Please update the commit message for the description to be at the top with more detail on what was fixed.

ch-stark:
no approved

dhaiducek:
@ch-stark The "not approved" is expected because you're not an approver in the OWNERS file: https://github.com/open-cluster-management-io/policy-collection/blob/main/OWNERS

But also, I think we should not be setting anything to `remediationAction: enforce` in the policy-collection. That should be left to the user. You could put `remediationAction: inform` with a comment there instead?

dhaiducek:
> But also, I think we should not be setting anything to `remediationAction: enforce` in the policy-collection. That should be left to the user. You could put `remediationAction: inform` with a comment there instead?

Oh, nevermind--I see it was already set to `enforce` and it was fine-tuned. 🙂

PR #439, opened 2023-12-13
Update ACS PolicySet to replace expired init-bundle

brian-jarvis:
Updates to existing stable ACS PolicySet for  OpenShift+ includes:

- Replacing Job with v2 as Job definition is immutable.  Removes original Job if it exists
- Creating policy to remove init-bundle created secrets and v2 Job if CertificatePolicy is in NotCompliant state.  Set logner evaluation period to give time for init-bundle Job to recreate and move CertificatePolicy to Compliant state.
- Corrected Policy dependencies such that CertificatePolicy is dependent on init-bundle creation and not sync of secrets. Removal of init-bundle output breaks sync policy until they are recreated
- Corrected Policy dependencies such that secret sync is dependent on init-bundle creation and not central install

Refs:

- https://issues.redhat.com/browse/ACM-9075


gparvin:
I added a hold while I test this.  I do see a Pending policy which means I need to update the OPP QE script that tests this policy set.  We just made an update last week to validate the policies aren't in the pending state :smile:

brian-jarvis:
The side effect of using a NonCompliant dependency, it will always show as Pending.

I was just talking with cstark about this earlier.  The policy showing as "pending" seems wrong to me.  Yes technically the dependency is not met and it is pending to execute.  But in this case it is expected to be in that state.  The policy executing means another Policy is in an out of compliance state.  If out of compliance is a red (bad) icon, then the dependency being satisfied should also be reflected as bad.

The overview page for Governance calls out pending policies as if something is incorrect and needs attention.  But in this use case that assumption seems incorrect since the policy is in the expected state.

gparvin:
We have an `ignorePending=true` flag that can be added which I think is what we want here.  That effectively keeps the policy in a Pending mode but the state is reported as Compliant.  So far it seems to be working anyway...

brian-jarvis:
@gparvin   I added `ignorePending`

gparvin (approved):
Thanks so much Brian!  This looks good to me and I love the new contribution.

PR #441, opened 2024-01-03
Bump versions in the validation script

dhaiducek:
This also dumps out deprecation warnings about the `commonLabels` usage, but that's a task for another day. 🙂


gparvin (approved):
Thanks!

PR #444, opened 2024-01-18
use copyMethod: Snapshot for ReplicationDestination

birsanv:
Update ReplicationDestination to use `copyMethod: Snapshot` 
This will force the applications using the PVC to wait until the restore completes, before being able to use the PVC


gparvin (approved):
Thanks for keeping things up to date.  This looks good to me.

PR #445, opened 2024-01-19
add pvc label to the pvc config

birsanv:
https://issues.redhat.com/browse/ACM-8622

Changes:
- add cluster.open-cluster-management.io/volsync-pvc: {{ $pvc.metadata.name }} to the configmap created by the replication source. It will be used on the restore hub , by the hub backup to wait on these pvcs to be created, after credentials backup is restored, and before restoring any other data
- updated the replication destination templated to remove the check for a PVC being bound; with the use of a ReplicationDestination of type copyMethod: Snapshot, the PVC will not be bound until the snapshot is completed and a user up bounds to it. So the check is not relevant and doesn't mark anything wrong with the replication
- replication destination template creating the resource, check for the existence of creds-time or creds-time-active restore. Depending on the type of restore ( sync versus all in once ) the creds-time-active  restore may be producing the PV configs


gparvin (approved):

birsanv:
@gparvin can you please help with the review process .. thank you

mprahl (approved):

gparvin (approved):

PR #449, opened 2024-01-26
Enable the quay bridge operator by default in the opp policyset

gparvin:
This sets up and configures the quay bridge operator by default in the opp policysets.  The quay bridge operator is deployed to all openshift managed clusters and will trust the default cert from the hub for the image registry.

This will allow a cross product interop test for OPP since now an application can be deployed that uses quay's image registry which is on ODF. ACS will scan the registry.  ACM will manage the app deployment.


gparvin:
/hold
Doing another test.

gparvin:
/unhold
I found a scenario that is working.

yiraeChristineKim (approved):

PR #453, opened 2024-02-03
Gatekeeper policy to replace deprecated IAMPolicy Controller

brian-jarvis:
This Gatekeeper Policy is intended to match the behavior of the deprecated ACM IAMPolicy Controller.  

It will allow an administrator to monitor and alert if `ClusterRoleBindings` with the specified `ClusterRole` exceed the maximum number of users.  In the case where a Group is specified in the `ClusterRoleBinding` the number of users in the group are counted.  ServiceAccounts are ignored.


community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
```diff
@@ -0,0 +1,262 @@
+
+# ## Introduction
+# This Gatekeeper Policy is intended to match the behavior of the deprecated ACM IAMPolicy Controller.  It will allow an administrator to monitor and alert if `ClusterRoleBindings` with the specified `ClusterRole` exceed the maximum number of users.  In the case where a Group is specified in the `ClusterRoleBinding` the number of users in the group are counted.  ServiceAccounts are ignored.
+#
+# ## Prerequisites
```
gparvin, line 5: We are delivering a new feature in the Gatekeeper operator which will automatically create the sync data for you. If you would like to try it out, you should be able to deploy the community gatekeeper operator policy and change the version to 3.14.0. FYI: @yiraeChristineKim

community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
```diff
@@ -0,0 +1,262 @@
...
+              constraint_kind: MaxIAMClusterBindings
+              constraint_name: max-cluster-admins
+              event_type: violation
+            apiVersion: v1
+            kind: Event
+        remediationAction: inform
+        severity: low
+---
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Placement
```
gparvin, line 233: Thanks for using placement instead of placementrule.  For new policies we are planning on not delivering the placement or placementbinding in the donated policy.  Curious if you have feedback regarding that too.

community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
```diff
@@ -0,0 +1,262 @@
+
+# ## Introduction
+# This Gatekeeper Policy is intended to match the behavior of the deprecated ACM IAMPolicy Controller.  It will allow an administrator to monitor and alert if `ClusterRoleBindings` with the specified `ClusterRole` exceed the maximum number of users.  In the case where a Group is specified in the `ClusterRoleBinding` the number of users in the group are counted.  ServiceAccounts are ignored.
```
gparvin, line 3: This is great info for the README.md too.

community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
brian-jarvis: I  will check it out, thanks for pointing that out.  Guess that means I won't need this policy to create the sync config from the annotations  https://github.com/bry-tam/acm-policy-samples/blob/main/policies/operators/gatekeeper/sync-configmap.yml

community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
brian-jarvis: I think not delivering the placement is the correct way to go.  The placements make it feel like it is a complete working solution when the user would still need to create the namespace binding.  I think we see a lot of issues pop up from that binding missing.

community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
brian-jarvis: Added.

gparvin:
/cc @dhaiducek @yiraeChristineKim 
The rego looks ok to me but thought you would be interested.

community/AC-Access-Control/policy-gatekeeper-limitclusteradmin.yaml
gparvin: Thanks for the feedback.  Can you confirm you intended to say you think we should continue to provide the binding and a placement?

brian-jarvis: I agree to the plan to not deliver the placement and binding.  I left them here since that plan is not final.  I can remove them if we want to start now with not delivering them.  I was expecting you would want to have a formal notice and documentation before it was implemented.

gparvin (approved):

yiraeChristineKim:
To me looks great!! wow, curious Do a lot of customers want to limit clusterbinding numbers?

PR #463, opened 2024-03-26
Added AMI ids to the config map for ocp 4.16

gparvin:
These are AMI ids I got from a 4.15 install which should still work fine for the 4.16 builds.


gparvin:
I tested the us-east-2 id so I know it should work.  The id for us-east-1 I got from the daily cluster master node which I think should still work fine.

yiraeChristineKim (approved):

PR #464, opened 2024-04-04
Fix the PolicyGenerator ArgoCD Policy

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-10650

### DIFF
--- a/community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
+++ b/community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
@@ -36,15 +36,14 @@ spec:
                     initContainers:
                     - args:
                       - -c
-                      - cp /etc/kustomize/plugin/policy.open-cluster-management.io/v1/policygenerator/PolicyGenerator
-                        /policy-generator/PolicyGenerator
+                      - cp /policy-generator/PolicyGenerator-not-fips-compliant /policy-generator-tmp/PolicyGenerator
                       command:
                       - /bin/bash
-                      image: 'registry.redhat.io/rhacm2/multicluster-operators-subscription-rhel8:v{{ (lookup "operator.open-cluster-management.io/v1" "MultiClusterHub" "open-cluster-management" "multiclusterhub").status.currentVersion }}'
+                      image: '{{ (index (lookup "apps/v1" "Deployment" "open-cluster-management" "multicluster-operators-hub-subscription").spec.template.spec.containers 0).image }}'
                       name: policy-generator-install
                       volumeMounts:
                       - mountPath: /policy-generator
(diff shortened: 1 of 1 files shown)


community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
```diff
@@ -36,15 +36,14 @@ spec:
                     initContainers:
                     - args:
                       - -c
-                      - cp /etc/kustomize/plugin/policy.open-cluster-management.io/v1/policygenerator/PolicyGenerator
-                        /policy-generator/PolicyGenerator
+                      - cp /policy-generator/PolicyGenerator-not-fips-compliant /policy-generator-tmp/PolicyGenerator
                       command:
                       - /bin/bash
-                      image: 'registry.redhat.io/rhacm2/multicluster-operators-subscription-rhel8:v{{ (lookup "operator.open-cluster-management.io/v1" "MultiClusterHub" "open-cluster-management" "multiclusterhub").status.currentVersion }}'
+                      image: '{{ (index (lookup "apps/v1" "Deployment" "open-cluster-management" "multicluster-operators-hub-subscription").spec.template.spec.containers 0).image }}'
```
gparvin, line 42: Is there a way to make it work if ACM isn't deployed to open-cluster-management namespace?

community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
```diff
@@ -36,15 +36,14 @@ spec:
                     initContainers:
                     - args:
                       - -c
-                      - cp /etc/kustomize/plugin/policy.open-cluster-management.io/v1/policygenerator/PolicyGenerator
-                        /policy-generator/PolicyGenerator
+                      - cp /policy-generator/PolicyGenerator-not-fips-compliant /policy-generator-tmp/PolicyGenerator
```
gparvin, line 39: Does the not fips compliant image exist yet?

community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
dhaiducek: Hmmm. Interesting idea--I suppose we could do a `lookup` using a label on the Deployment. That should only return one result that we can parse...

community/CM-Configuration-Management/policy-openshift-gitops-policygenerator.yaml
dhaiducek: Ah, good point. Though we can merge this now and wait to cherry-pick it to 2.10.

yiraeChristineKim (approved):

PR #465, opened 2024-04-16
Remove `ManagedClusterConditionAvailable`

dhaiducek:
Setting `ManagedClusterConditionAvailable` in a `PlacementRule` is tumultuous in Policy if managed clusters become unavailable because the Policy will be removed from the managed cluster.

### DIFF
--- a/community/CM-Configuration-Management/policy-oauth-ldapsync.yaml
+++ b/community/CM-Configuration-Management/policy-oauth-ldapsync.yaml
@@ -298,6 +298,3 @@ spec:
         operator: In
         values:
           - dev
-  clusterConditions:
-  - status: "True"
-    type: ManagedClusterConditionAvailable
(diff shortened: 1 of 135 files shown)


dhaiducek:
(This PR is large because I used `yq`, so it reformatted files that didn't follow its conventions.)

community/CM-Configuration-Management/policy-oauth-ldapsync.yaml
```diff
@@ -109,7 +109,7 @@ spec:
               objectDefinition:
                 apiVersion: v1
                 data:
-                  ldap-sync.yml: >
+                  ldap-sync.yml: |
```
mprahl: I'm wondering why it changed this string. I'd prefer to revert this plus the random new line after `augmentedActiveDirectory`.

dhaiducek: That was me. The underlying `go-yaml` parser in `yq` translates multiline strings to show as they would after the multiline marker is applied, so I fiddled with it to get it to show properly, but I forgot to flip it back. I'll revert it.

mprahl (approved):

mprahl:
Thanks for doing this @dhaiducek!

PR #476, opened 2024-05-09
Replace odf community policy with an improved sample

gparvin:
Community policies are not supported, but we want our samples to be useful.  Ideally we could run tests or some checks to determine what policies are out of date or no longer applicable, but for this policy I am updating it from the OPP policy set contents which deploys ODF.

Refs:
 - https://issues.redhat.com/browse/ACM-10092

### DIFF
--- a/community/CM-Configuration-Management/policy-odf.yaml
+++ b/community/CM-Configuration-Management/policy-odf.yaml
@@ -2,9 +2,6 @@
 # the OpenShift Data Foundation on the managed clusters.
 #
 # If set to "enforce" it'll install the operator.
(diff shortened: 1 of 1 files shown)


community/CM-Configuration-Management/policy-odf.yaml
```diff
@@ -35,7 +32,7 @@ spec:
         apiVersion: policy.open-cluster-management.io/v1
         kind: ConfigurationPolicy
```
mprahl, line 53: Could this be changed to OperatorPolicy?

gparvin: Yes, I updated the policy.  See what you think.

mprahl (approved):

PR #478, opened 2024-05-22
Add autoimport policy to include MultiClusterEngineHCP

serngawy:
This policy to allow auto import clusters with type MultiClusterEngineHCP from discovered clusters

### DIFF
--- a/community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
+++ b/community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
@@ -1,16 +1,17 @@
 apiVersion: policy.open-cluster-management.io/v1
 kind: Policy
 metadata:
-  name: policy-rosa-autoimport
+  name: policy-mce-hcp-autoimport
+  namespace: open-cluster-management-global-set
   annotations:
     policy.open-cluster-management.io/standards: NIST SP 800-53
     policy.open-cluster-management.io/categories: CM Configuration Management
     policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
     policy.open-cluster-management.io/description: Discovered clusters that are of
-      type ROSA can be automatically imported into ACM as managed clusters.  This
-      policy helps you select those managed clusters and configure them so the import
-      will happen.  If you do not want all of your ROSA clusters to be automatically
-      imported, you can configure filters or add an annotation.
+      type MultiClusterEngineHCP can be automatically imported into ACM as managed clusters.
+      This policy helps you select those managed clusters and configure them so the import
+      will happen. Fine tuning MultiClusterEngineHCP clusters to be automatically imported
+      can be done by configure filters at the configMap or add annotation to the discoverd cluster.
 spec:
   # Remove the default remediation below to enforce the policies.
   remediationAction: inform
@@ -20,7 +21,7 @@ spec:
         apiVersion: policy.open-cluster-management.io/v1
         kind: ConfigurationPolicy
         metadata:
-          name: rosa-autoimport-config
+          name: mce-hcp-autoimport-config
         spec:
           object-templates:
             - complianceType: musthave
@@ -38,12 +39,12 @@ spec:
         apiVersion: policy.open-cluster-management.io/v1
         kind: ConfigurationPolicy
         metadata:
-          name: policy-rosa-autoimport
+          name: policy-mce-hcp-autoimport
         spec:
           remediationAction: enforce
           severity: low
           object-templates-raw: |
-            {{- /* find the ROSA DiscoveredClusters */ -}}
+            {{- /* find the MultiClusterEngineHCP DiscoveredClusters */ -}}
             {{- range $dc := (lookup "discovery.open-cluster-management.io/v1" "DiscoveredCluster" "" "").items }}
               {{- /* Check for the flag that indicates the import should be skipped */ -}}
               {{- $skip := "false" -}}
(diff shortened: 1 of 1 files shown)


community/CM-Configuration-Management/policy-rosa-autoimport.yaml
```diff
@@ -98,4 +98,3 @@ spec:
                       status: "True"
               {{- end }}
             {{- end }}
-
```
gparvin, line 101: Accidental change since I see you've split these 2 policies.

community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
```diff
@@ -0,0 +1,100 @@
...
+                  name: discovery-config
+                  namespace: open-cluster-management-global-set
+                data:
+                  mce-hcp-filter: ""
+            {{- /* find the MultiClusterEngineHCP DiscoveredClusters */ -}}
+            {{- range $dc := (lookup "discovery.open-cluster-management.io/v1" "DiscoveredCluster" "" "").items }}
+              {{- /* Check for the flag that indicates the import should be skipped */ -}}
+              {{- $skip := "false" -}}
+              {{- range $key, $value := $dc.metadata.annotations }}
+                {{- if and (eq $key "discovery.open-cluster-management.io/previously-auto-imported")
```
gparvin, line 52: I'll check with Disaiah to make sure this annotation will work for any type.

community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
```diff
@@ -0,0 +1,100 @@
...
+          object-templates-raw: |
+            - complianceType: musthave
+              objectDefinition:
+                apiVersion: v1
+                kind: ConfigMap
+                metadata:
+                  name: discovery-config
+                  namespace: open-cluster-management-global-set
+                data:
+                  mce-hcp-filter: ""
```
gparvin: I had to split the configmap out into a separate ConfigurationPolicy in PR https://github.com/open-cluster-management-io/policy-collection/pull/479.  I think you will need to also.  Likely what you hit yesterday when you mentioned you were having trouble with the configmap which I didn't realize at the time.

community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
```diff
@@ -0,0 +1,100 @@
...
+              {{- /* Check for the flag that indicates the import should be skipped */ -}}
+              {{- $skip := "false" -}}
+              {{- range $key, $value := $dc.metadata.annotations }}
+                {{- if and (eq $key "discovery.open-cluster-management.io/previously-auto-imported")
+                           (eq $value "true") }}
+                  {{- $skip = "true" }}
+                {{- end }}
+              {{- end }}
+              {{- /* if the type is MultiClusterEngineHCP and the status is Active */ -}}
+              {{- if and (eq $dc.spec.status "Active")
```
gparvin, line 58: The DiscoveredCluster CRD expects the status field to be set, but I'm not sure if it is set the same way for MCE.  Just an FYI -- maybe it will always be active for this type.

community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
gparvin: This annotation will work for other types.  It may not yet, Disaiah mentioned a PR that may be needed.

serngawy: I test it, it works

community/CM-Configuration-Management/policy-mce-hcp-autoimport.yaml
serngawy: it act the same as ROSA

gparvin (commented):
I think your policy looks good.  I see your commit is still trying to change the ROSA policy too, I think those changes are just reflecting the already delivered changes, but wondering if you can remove that file from the commit.

serngawy:
> I think your policy looks good. I see your commit is still trying to change the ROSA policy too, I think those changes are just reflecting the already delivered changes, but wondering if you can remove that file from the commit.

I made some changes to the ROSA policy as well like the condition type wasn't correct [here](https://github.com/open-cluster-management-io/policy-collection/pull/478/files#diff-d21df5d0c1ae2d6d5d75b4da81375b8a9ed1e64befde271c8351946c889833f5L97)  and removing the global remediation [here](https://github.com/open-cluster-management-io/policy-collection/pull/478/files#diff-d21df5d0c1ae2d6d5d75b4da81375b8a9ed1e64befde271c8351946c889833f5L15)  The other changes I guess github missing around.

community/CM-Configuration-Management/policy-rosa-autoimport.yaml
```diff
@@ -12,18 +12,15 @@ metadata:
       will happen.  If you do not want all of your ROSA clusters to be automatically
       imported, you can configure filters or add an annotation.
 spec:
-  remediationAction: inform
```
gparvin, line 15: By convention we want to make policies inform by default.  This way the customer has an opportunity to see what resources the policy will apply to before enforcing.  For this policy I think the inform is useful so the customer can determine if they want to use a filter to target specific Discovered clusters or if they are ok with all that are matched.

serngawy: hmm, it's just confused me as the configuration-policy "policy-rosa-managedcluster-status" is inform and the reset are enforce. okay, I will put it back.

gparvin (approved):
Thanks!  This looks good and appreciate the fix too!

PR #480, opened 2024-05-23
Policy configmap creation was blocked by template function

gparvin:
The configmap can't be created due to a template not resolving and the template won't resolve because the configmap didn't exist.  I likely had the configmap laying around while testing and didn't do a clean run


gparvin:
The issue for this is: https://issues.redhat.com/browse/ACM-11848

gparvin:
I created the PR twice so closing this one.

PR #483, opened 2024-06-07
Adjust machine sets to work on ocp 4.16

gparvin:
The machineset no longer works on ocp 4.16.  This pr adds in some filters which seem to be additional required configuration that resolves the issue.

Refs:
 - https://issues.redhat.com/browse/OCPQE-22447

### DIFF
--- a/policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
+++ b/policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
@@ -77,6 +77,17 @@ spec:
                     - name: tag:Name
                       values:
                       - '{{ (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.infrastructureName }}-worker-sg'
+                  {{- $ocpversion := semver (fromClusterClaim "version.openshift.io") }}
+                  {{- if gt $ocpversion.Minor 15 }}
+                  - filters:
+                    - name: tag:Name
+                      values:
+                      - '{{ (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.infrastructureName }}-node'
+                  - filters:
+                    - name: tag:Name
+                      values:
+                      - '{{ (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.infrastructureName }}-lb'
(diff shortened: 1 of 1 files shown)


gparvin:
Would like a review before end of day Tuesday if possible.  Thanks!

policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
```diff
@@ -77,6 +77,17 @@ spec:
                     - name: tag:Name
                       values:
                       - '{{ (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.infrastructureName }}-worker-sg'
+                  {{- $ocpversion := semver (fromClusterClaim "version.openshift.io") }}
+                  {{- if gt $ocpversion.Minor 15 }}
```
dhaiducek, line 81: We have `semverCompare` available--maybe use it directly?
```suggestion
                  {{- semverCompare ">4.15" (fromClusterClaim "version.openshift.io") }}
```

gparvin: I saw the semverCompare function, but I wasn't sure how it handled the patch release if you did a compare as described.  I didn't play with it -- just went down the path that made sense at the moment.  I'll test this some and see what happens.

dhaiducek: Oh, you're probably right. It probably should be `">=4.16"`

gparvin: When I use any of the comparisons with 4.16.0-rc.2 builds the comparisons just don't seem to work.  While it works fine if I change the comparison to run against just "4.16.0", I think I am leaning towards keeping the existing code rather than switch to semverCompare which appears to have trouble with some OCP versioning - even if these are rc builds.

dhaiducek: Works for me! And I guess the risk of bumping to OCP v5 is fairly low right now 🙂

dhaiducek (approved):

PR #498, opened 2024-08-19
Use local-cluster label

JustinKuli:
For a while now, self-managed hub ManagedCluster resources have had the `local-cluster='true'` label, and we may start seeing those clusters have different names. This PR adjusts many occurrences in the collection of using the name; now they use the label.

The change to the backup policy to avoid a named lookup is in a [separate commit](https://github.com/JustinKuli/policy-collection/commit/605888411104c3deb0d66b75de1ae18c3531867e) because I think it needs extra attention, it's less obvious than the other changes.


dhaiducek (approved):
/hold I think @birsanv should give it a quick review also since we're modifying the backup/restore policy.

Nice work!

birsanv:
thank you for the changes @JustinKuli ! 

/lgtm

PR #500, opened 2024-09-06
Add policy for ArgoCD healthcheck configuration

JustinKuli:
This uses object-templates-raw and a ranged lookup to configure *all* existing ArgoCD instances, which also has the beneficial side-effect of not incorrectly creating the ArgoCD instance before the operator has had a chance to set it up with an initial configuration.


stable/CM-Configuration-Management/argocd-policy-healthchecks.yaml
```diff
@@ -0,0 +1,162 @@
...
+                          hs.status = "Progressing"
+                          hs.message = "Waiting for the status to be updated"
+                          return hs  
+                        end
+                        if obj.status.compliant == "Compliant" then
+                          hs.status = "Healthy"
+                        else
+                          hs.status = "Degraded"
+                        end
+                        if obj.status.compliancyDetails ~= nil then
```
yiraeChristineKim, line 75: Is it guaranteed that the flow will reach this point and return, regardless of whether obj.status.compliant == "Compliant" or not? I’m asking because hs.status = "Progressing" is being overridden further down.

JustinKuli: Good question. It would have been more clear if I had written it as:
```lua
if obj.status.compliancyDetails == nil then
  hs.status = "Progressing"
  hs.message = "Waiting for compliance"
  return hs
end
```
and I could have put that above the `if obj.status.compliant == "Compliant"` part.

These checks are already present in the argocd repo, and I'd like to keep them in sync. Since this change would be logically equivalent, I'd rather just keep it as is for now, what do you think?

mprahl (approved):

PR #502, opened 2024-09-25
check for completed backups and update compliant messages (https://issues.redhat.com/browse/ACM-14460)

birsanv:
https://issues.redhat.com/browse/ACM-14460

Changes:
- Updated the check-backup-completed policy template to look for any backup in completed state, not only for the latest backup.
- The other templates are still validating the latest backup phase ( will show violation if the Backup matching the Schedule.status.lastBackup  has a status of Error, FailedValidation or '')
- Updated compliance messages with a more readable content, see attached image
The reason for the change : sometimes - and it seems that it usually happens first time the Schedule gets created but cannot always be reproduced - the Schedule.status.lastBackup timestamp if off and does not match the generated Backup startTimestamp. 
Because of this we show a violation when we check if the latest backup is completed. The violation is because the Backup was not found based on the Schedule.status.lastBackup timestamp rule, but the Backup exists and is Completed. 
To avoid this false positive, I am just checking if there is a Backup generated by this Schedule, with a status of Completed. This check will show compliant though even if the Completed backup was generated long ago. This should be ok though because we have other templates checking  that the latest backup must not have a status of ( Error, FailedValidation or '' ). So if those are compliant, the latest backup should be completed

Screenshot 2024-09-25 at 11 17 12 AM

### DIFF
--- a/community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-backup.yaml
+++ b/community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-backup.yaml
@@ -78,29 +78,12 @@ spec:
             {{hub end hub}}
           remediationAction: inform
           severity: high
-    - objectDefinition:
-        apiVersion: policy.open-cluster-management.io/v1
-        kind: ConfigurationPolicy
-        metadata:
-          name: check-backup-completed
-        spec:
-          object-templates-raw: |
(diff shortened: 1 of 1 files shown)


sahare (approved):
Looks good, Thanks!

gparvin (approved):

JustinKuli:
Very clever use of hub-templates to get `lookup` calls in the `customMessage` fields!

PR #503, opened 2024-10-01
Change the OPP policies to adopt OperatorPolicy

gparvin:
Deploying operators with OperatorPolicy has benefits that should be adopted to help customers get better visibility into the operator status automatically.

Refs:
 - https://issues.redhat.com/browse/OCPQE-25041

### DIFF
--- a/policygenerator/policy-sets/stable/openshift-plus/input-acs-central/policy-acs-operator-central.yaml
+++ b/policygenerator/policy-sets/stable/openshift-plus/input-acs-central/policy-acs-operator-central.yaml
@@ -8,24 +8,24 @@ kind: Namespace
 metadata:
   name: rhacs-operator
 ---
-apiVersion: operators.coreos.com/v1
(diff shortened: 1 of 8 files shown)


mprahl:
@gparvin I'd recommend setting the channels explicitly for best compatibility. We've encountered issues with autodetecting the channel if the cluster has multiple catalogs providing the same operator. This is because the OLM `PackageManifest` API round robins between them in some unknown way.

gparvin:
> @gparvin I'd recommend setting the channels explicitly for best compatibility. We've encountered issues with autodetecting the channel if the cluster has multiple catalogs providing the same operator. This is because the OLM `PackageManifest` API round robins between them in some unknown way.

@mprahl I took your recommendation for a couple operators, but the others don't have a channel name that I can reliably generate without having to manually track it.  ODF and Quay had versions built in their channel names, example: `stable-3.12`.

mprahl (approved):

PR #510, opened 2024-12-10
Configure Compliance Operator for ROSA HCP

JustinKuli:
As mentioned in the [Compliance Operator documentation](https://docs.openshift.com/container-platform/4.17/security/compliance_operator/co-management/compliance-operator-installation.html#installing-compliance-operator-rosa_compliance-operator-installation), when installed on ROSA hosted control planes, a special node selector must be configured for the operator to run. This update uses conditionals to only make this configuration only on those clusters.

Refs:
 - https://issues.redhat.com/browse/ACM-14161

### DIFF
--- a/stable/CA-Security-Assessment-and-Authorization/policy-compliance-operator-install.yaml
+++ b/stable/CA-Security-Assessment-and-Authorization/policy-compliance-operator-install.yaml
@@ -51,3 +51,9 @@ spec:
             namespace: openshift-compliance
             source: redhat-operators
             sourceNamespace: openshift-marketplace
+            # Conditionally configure a nodeSelector for installing on ROSA hosted control planes
+            config: '{{ if and (eq "ROSA" (fromClusterClaim
+              "product.open-cluster-management.io")) (eq "true"
+              (fromClusterClaim "hostedcluster.hypershift.openshift.io"))
+              }}{"nodeSelector":{"node-role.kubernetes.io/worker":""} }{{ else
+              }}{{ "{}" | toLiteral }}{{ end }}'


yiraeChristineKim (approved):

stable/CA-Security-Assessment-and-Authorization/policy-compliance-operator-install.yaml
```diff
@@ -51,3 +51,9 @@ spec:
             namespace: openshift-compliance
             source: redhat-operators
             sourceNamespace: openshift-marketplace
+            # Conditionally configure a nodeSelector for installing on ROSA hosted control planes
+            config: '{{ if and (eq "ROSA" (fromClusterClaim
```
dhaiducek, line 55: Nice work! Not necessary to update now, but this would be cleaner as a folded multiline string:
```
config: >-
  {{ if and 
    (eq "ROSA" (fromClusterClaim "product.open-cluster-management.io"))
    (eq "true" (fromClusterClaim "hostedcluster.hypershift.openshift.io"))
  }}
    {"nodeSelector":{"node-role.kubernetes.io/worker":""} }
  {{ else }}
    {{ "{}" | toLiteral }}
  {{ end }}
```

~But actually, now that I do that, it looks like it'd fail--it has an extra `}` in there after the nodeSelector?~

JustinKuli: Using `>-` is generally a good idea! But multiline yaml can be incredibly frustrating. In this case, the indents break the line folding, so I think this would be necessary:
```yaml
config: >-
  {{ if and
  (eq "ROSA" (fromClusterClaim "product.open-cluster-management.io"))
  (eq "true" (fromClusterClaim "hostedcluster.hypershift.openshift.io"))
  }}
  {"nodeSelector":{"node-role.kubernetes.io/worker":""} }
  {{ else }}
  {{ "{}" | toLiteral }}
  {{ end }}
```

You also had a space at the end of the first line, emphasizing how tricky this is.

PR #512, opened 2025-02-13
Refresh the AMIs and add ones for ocp 4.19

gparvin:
The AMI IDs need to be updated occassionally and when we want to add support for a new ocp release.  Adding in 4.19 ami ids.


dhaiducek (approved):
/hold LGTM! But in case you want to address either of these:

- I know I found `https://github.com/openshift/installer/blob/release-<version>/data/data/coreos/rhcos.json` last time--is that what you used this time also, and should we document it (or create a script to keep them up-to-date)?
- I noticed you only updated 4.16+ -- Should we drop lower versions? I wasn't sure if 4.12 was still in support...

gparvin:
Looks like I did this in another pr already.

PR #513, opened 2025-03-28
Refresh the AMI ids used when adding nodes to OPP setup for storage

gparvin:
The OPP policy setup requires storage nodes to be provisioned.  The machinesets defined need to have AMI ids for the AWS nodes that need to be added for the storage nodes.  Adding OCP 4.19 AMI IDs and refreshing the other IDs to keep the node OS up to date.

### DIFF
--- a/policygenerator/policy-sets/community/openshift-plus-setup/opp-settings.yaml
+++ b/policygenerator/policy-sets/community/openshift-plus-setup/opp-settings.yaml
@@ -17,12 +17,14 @@ data:
   us-east-2-4.14: ami-0dd810c1f47c5c233
   us-east-1-4.15: ami-0d653d86d4113326a
   us-east-2-4.15: ami-0d6c4efce8daf7d2d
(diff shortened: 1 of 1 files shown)


dhaiducek:
@gparvin I ran the script and those last two appear to be labeled as RHEL CoreOS?

```
OCP-4.19	us-east-1	RHEL CoreOS 9.6 9.6.20250321-0 x86_64
OCP-4.19	us-east-2	RHEL CoreOS 9.6 9.6.20250321-0 x86_64
```

Is that expected?

dhaiducek (approved):
Ah, I guess the AMI is just the machine to install Openshift on top of, so perhaps the labels are just more accurate now. 🙂

gparvin:
> Ah, I guess the AMI is just the machine to install Openshift on top of, so perhaps the labels are just more accurate now. 🙂

Yes, that's what I'm hoping! Thanks!

PR #514, opened 2025-04-08
Add standalone-hub-templates example

JustinKuli:
An example of enabling and configuring details for the standalone-hub-templating feature.


gparvin (approved):
Thanks for creating a sample!

PR #517, opened 2025-05-09
Add scanner v4 config

shaneboulden:
Adds config for RHACS Scanner V4, though does not enable it per the current RHACS 4.7.2. default config.

Scanner V4 can be enabled by changing the `scannerComponent` to `Enabled`, and optionally disabling the StackRox Scanner.

### DIFF
--- a/policygenerator/policy-sets/stable/openshift-plus/input-acs-central/policy-acs-operator-central.yaml
+++ b/policygenerator/policy-sets/stable/openshift-plus/input-acs-central/policy-acs-operator-central.yaml
@@ -47,6 +47,24 @@ spec:
         claimName: stackrox-db
   egress:
     connectivityPolicy: Online
+  scannerV4:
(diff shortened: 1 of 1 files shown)


gparvin:
Thanks for the PR.  I tried it out and it seems to work fine.  If we disable the current scanner there is a status check that we would need to update but for now I think this is great.
It looks like the commit is missing the DCO signoff.  Since there's only one commit here, you can add it using the --amend flag and a force-push:

```
git commit --amend --signoff --no-edit
git push --force
```

Git doc ref: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s

shaneboulden:
Thanks @gparvin, just added the DCO signoff.

gparvin (approved):
Thanks, I forgot about this while on vacation.  I appreciate the fix!

PR #521, opened 2025-06-10
add oadp operator startingCSV

birsanv:
https://issues.redhat.com/browse/ACM-21326

Add support for setting optional OADP operator settings, such as starting CSV

New options for the hdr-app-configmap ConfigMap ( and channel made as optional ) :
These are documented in README

hdr-app-configmap ConfigMap               | Optional |    Default Value           | Description            
-------------------------------------------| ----------------------- | ------   |-----------------------------
channel                                    | Yes | <i>stable</i> for OCP 4.19 or higher, <i>stable-1.4</i> otherwise  | Used by the [oadp-hdr-app-install](./resources/policies/oadp-hdr-app-install.yaml) policy and sets the OADP operator install channel. OADP channel version defaults to the OADP supported version based on the cluster OCP version. Use this property to overwrite the default value.
packageName                                  | Yes |  <i>redhat-oadp-operator</i>   | Used by the [oadp-hdr-app-install](./resources/policies/oadp-hdr-app-install.yaml) policy and sets OADP channel name. Set this property for custom installation, for example: offline install.
subscriptionSource                           | Yes | <i>redhat-operators</i>  | Used by the [oadp-hdr-app-install](./resources/policies/oadp-hdr-app-install.yaml) policy and sets OADP subscription. Set this property for custom installation, for example: offline install.
subscriptionSourceNamespace                  | Yes | <i>openshift-marketplace</i>  | Used by the [oadp-hdr-app-install](./resources/policies/oadp-hdr-app-install.yaml) policy and sets OADP subscription source. Set this property for custom installation, for example: offline install.
subscriptionStartingCSV                      | Yes | None  | Used by the [oadp-hdr-app-install](./resources/policies/oadp-hdr-app-install.yaml) policy and sets OADP subscription startingCSV.
subscriptionInstallPlanApproval              | Yes | <i>Automatic</i>   | Used by the [oadp-hdr-app-install](./resources/policies/oadp-hdr-app-install.yaml) policy and sets OADP install plan.


birsanv:
@gparvin can I please have a review on this

birsanv:
hi @gparvin @JustinKuli can I please have a review for this PR

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
```diff
@@ -76,21 +76,84 @@ spec:
...
-            {{hub with $configMap := (lookup "v1" "ConfigMap" "" "hdr-app-configmap") hub}}
+            {{- /* inferre the oadp channel from the cluster ocp version */ -}}
+            {{- $oadp_channel := "stable-1.4" }}
+            {{- $ocp_version := "" }}
+            {{- $cluster_version := (lookup "config.openshift.io/v1" "ClusterVersion" "" "version")  }}
+            {{- if eq $cluster_version.metadata.name  "version" }}
+              {{- range $hist := $cluster_version.status.history}}
+                {{- $ocp_version = $hist.version }}
+              {{- end }}
+            {{- end }}
```
JustinKuli: I suspect there is a `ClusterClaim` or something in the `.ManagedClusterLabels` for this - do those not work?

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
```diff
@@ -76,21 +76,84 @@ spec:
...
+            {{ if or (hasPrefix "4.19" $ocp_version) (hasPrefix "4.2" $ocp_version) }}
+            {{- $oadp_channel = "stable" }}
+            {{- end }}
```
JustinKuli: Presumably at some point in the next 9 OCP versions, this will need to be updated again.

Have you tried OperatorPolicy for this? If the `channel` field there is left empty, it will lookup the package from the catalog and can find the default channel there, which *can* be set up to vary based on OCP version. If the OADP operator doesn't do that, then I would accept this as being necessary. But I would rather avoid this huge amount of templating if possible.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
```diff
@@ -76,21 +76,84 @@ spec:
...
+            {{ if hasPrefix "4.12" $ocp_version }}
+            {{- $oadp_channel = "stable-1.3" }}
+            {{- end }}
+            {{- $subscriptionInstallPlanApproval := "Automatic" }}
+            {{- /* channelName - optional, set this property for custom installation only */ -}}
+            {{- $channelName := "redhat-oadp-operator" }}
```
JustinKuli: A more accurate name might be `packageName`

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
```diff
@@ -76,21 +76,84 @@ spec:
...
+                  {{- $subscription_name = $ss.metadata.name }}
+                  {{- $channelName = $ss.spec.name }}
+                  {{- $subscriptionInstallPlanApproval := $ss.spec.installPlanApproval }}
+                  {{- $subscriptionSource = $ss.spec.source }}
+                  {{- $subscriptionSourceNamespace = $ss.spec.sourceNamespace }}
+                {{- else }}
+                  {{- $found_other_version = true }}
+                {{- end }}
+              {{- end }}
+            {{- end }}
```
JustinKuli: I'm not sure this works as described, if I'm reading it correctly.

If another subscription for OADP is found, but it doesn't match the desired `oadp_channel` channel setting, then this sets `found_other_version = true` and doesn't change other variables. But the `objectDefinition` below will still be enforced, using the `oadp_channel` defined above.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
```diff
@@ -76,21 +76,84 @@ spec:
...
+              {{- end }}
+              {{- $subscriptionNsConfig := "{{hub fromConfigMap "" "hdr-app-configmap" "subscriptionSourceNamespace" hub}}" }}
+              {{ if not (eq "" $subscriptionNsConfig) }}
+                {{- $subscriptionSourceNamespace = $subscriptionNsConfig }}
+              {{- end }}
+              {{- $subscriptionSourceConfig := "{{hub fromConfigMap "" "hdr-app-configmap" "subscriptionSource" hub}}" }}
+              {{ if not (eq "" $subscriptionSourceConfig) }}
+                {{- $subscriptionSource = $subscriptionSourceConfig }}
+              {{- end }}
+              {{- $startingCSV := "{{hub fromConfigMap "" "hdr-app-configmap" "subscriptionStartingCSV" hub}}" }}
```
JustinKuli: 1. I'm worried about the mix of hub templates and managed cluster templates. It seems like an easy source for errors.
2. Is the idea that the users can modify the ConfigMap, instead of the policy? I've heard of that for things on each managed cluster, but this ConfigMap is on the hub.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
```diff
@@ -76,21 +76,84 @@ spec:
...
+            {{- $oadp_channel = "stable-1.3" }}
+            {{- end }}
+            {{- $subscriptionInstallPlanApproval := "Automatic" }}
+            {{- /* channelName - optional, set this property for custom installation only */ -}}
+            {{- $channelName := "redhat-oadp-operator" }}
+            {{- $subscriptionSourceNamespace := "openshift-marketplace" }}
+            {{- $subscriptionSource := "redhat-operators" }}
+
+            {{- /* use the subscription name and operator group from the existing OADP installation, if one exists in this ns  */ -}}
+            {{- $subscription_name := "redhat-oadp-operator-subscription" }}
```
JustinKuli: It's conventional for the subscription name to match the package name. Actually, I don't think it's possible to do it like this through the OCP UI (I could be wrong).

OperatorPolicy requires them to be the same, so if this is absolutely needed, I guess that would be a reason it can't be used here.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
JustinKuli: Also, if you want to handle an existing Subscription for the operator, you need to consider that the OperatorGroup might already exist. If the operator is installed through the UI, then the OperatorGroup has a randomly generated suffix. The ConfigurationPolicy here would create a new one, and I'd expect the Subscription would go into an error state.

Again, OperatorPolicy would help deal with this, if it could be used.

JustinKuli (changes requested):
This is a lot more complex templating than I'd like to see. I suppose I need some justification for why it's all needed.

Many of the tricky parts here are handled by OperatorPolicy. If you're not familiar, here are [the docs](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.13/html-single/governance/index#operator-policy-ctrl), and [a blog](https://developers.redhat.com/articles/2024/08/08/getting-started-operatorpolicy) to introduce it.

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
birsanv: You are right, I changed to use clusterClaim instead {{- $ocp_version := fromClusterClaim "version.openshift.io" }}

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
birsanv: I updated the rules to set the channel to stable if the ocp version is 4.19 or prefix not 4.1

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
birsanv: changed to use packageName

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
birsanv: I removed this completely, assuming the operator is always installed by the policy

community/CM-Configuration-Management/acm-app-pv-backup/resources/policies/oadp-hdr-app-install.yaml
birsanv: The reason for using hub template here is that the user defines the policy configuration by creating a configMap on the hub, under the policy namespace . The ConfigMap name is hdr-app-configmap
When the policy is applied on a managed clusters, it gets the defaults from the hdr-app-configmap on the hub. I could copy the ConfigMap on the managed cluster and read the data from the managed cluster but it works this way too.

birsanv:
hi @JustinKuli thanks for taking the time to review my changes, I really appreciate this and your comments
This policy is part of a 3 set policies ( one installing oadp , the other backing up data and the last one restore data ). I was trying to push some new features I have already implemented with the virt policies available with ACM 2.13 and your comments made me realize that I am doing it backwards ; I think I need to restructure these app policies, from the community project and make them reuse instead of copying the ACM virt policy support.
I'll close this PR and I'll work at the restructure part; basically this oadp-hdr-app-install is no longer needed because the oadp install can be done by the virt policy

PR #525, opened 2025-08-04
ODF 4.19 caused problems with the OPP PolicySet

gparvin:
On OCP 4.19, with the release of ODF 4.19, there is a problem that prevents OPP from installing successfully.  The ODF operator no longer provides the StorageSystem CRD.  In ODF 4.19 the customer is expected to create the StorageCluster without a StorageSystem.

Additionally, the StorageCluster could not be created as configured in ODF 4.19.  I had to scope the configuration of ODF back to a more basic install. I have tested the changes on OCP 4.19 and OCP 4.16 successfully.

Refs:
 - https://issues.redhat.com/browse/LPINTEROP-5783

### DIFF
--- a/policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
+++ b/policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
@@ -44,46 +44,12 @@ spec:
             cephConfig: {}
             cephDashboard: {}
             cephFilesystems: {}
+            cephNonResilientPools: {}
             cephObjectStoreUsers: {}
             cephObjectStores: {}
+            cephRBDMirror: {}
             cephToolbox: {}
-          mirroring: {}
-          nodeTopologies: {}
-          resources:
-            mds: {}
-            mgr: {}
-            mon: {}
-            noobaa-core: {}
-            noobaa-db: {}
-            noobaa-endpoint:
-              limits:
-                cpu: 1
-                memory: 500Mi
-              requests:
-                cpu: 1
-                memory: 500Mi
-            rgw: {}
-          storageDeviceSets:
-          - config: {}
-            count: 1
-            dataPVCTemplate:
-              metadata: {}
-              spec:
-                accessModes:
-                - ReadWriteOnce
-                resources:
-                  requests:
-                    storage: 100Gi
-    {{- if (eq (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").spec.platformSpec.type "VSphere") }}
-                storageClassName: thin-csi-odf
-    {{- else }}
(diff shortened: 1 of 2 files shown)


policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
```diff
@@ -44,46 +44,12 @@ spec:
...
-                volumeMode: Block
-              status: {}
-            name: ocs-deviceset
-            placement: {}
-            portable: true
-            preparePlacement: {}
-            replica: 3
-            resources: {}
+          multiCloudGateway:
+            dbStorageClassName: gp3-csi
```
JustinKuli, line 53: Previously, this was in a template:
```
    {{- if (eq (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").spec.platformSpec.type "VSphere") }}
                storageClassName: thin-csi-odf
    {{- else }}
                storageClassName: gp3-csi
    {{- end }}
```

Was that deliberately removed?

policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf.yaml
```diff
@@ -25,22 +25,32 @@ spec:
     source: redhat-operators
     sourceNamespace: openshift-marketplace
 ---
-apiVersion: odf.openshift.io/v1alpha1
-kind: StorageSystem
-metadata:
-  name: ocs-storagecluster-storagesystem
-  namespace: openshift-storage
-  labels:
-    removed-in-ocp419: '{{ skipObject (semverCompare  ">= 4.19" (fromClusterClaim "version.openshift.io")) }}'
```
JustinKuli, line 34: I'm surprised this didn't work! What went wrong with it? The `object-templates-raw` approach seems basically the same

JustinKuli (commented):
I just have questions to make sure I'm not missing something

policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
gparvin: No, let me add that back in if it still applies.

policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf.yaml
gparvin: The skipObject in the label was still trying to resolve the CRD.  I expect if the CRD was defined, maybe it would have worked as expected -- but because it was looking for the CRD I had to use a different solution.

JustinKuli (approved):
LGTM

PR #527, opened 2025-08-15
Switch back to the storage class that vsphere uses

gparvin:
Some changes recently to odf caused a storage class to be set to gp3-csi which needs to be thin-csi-odf for vsphere.


gparvin:
/cc @JustinKuli 
You noticed this concern in your last review and I think you were correct!

JustinKuli (approved):

PR #530, opened 2025-09-24
Changing the ODF configuration for OPP

gparvin:
This is switching the ODF configuration back to the storage device sets that we were using prior to ODF 4.19.  While this was causing problems it is the desired configuration from the ODF team and it appears some fixes on top of ODF 4.19 may have improved the experience.  I have deployed this successfully on OCP 4.19.

Refs:
 - https://steps.ci.openshift.org/reference/odf-apply-storage-cluster
 - https://issues.redhat.com/browse/LPINTEROP-5791

### DIFF
--- a/policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
+++ b/policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
@@ -31,26 +31,37 @@ spec:
           name: ocs-storagecluster
           namespace: openshift-storage
         spec:
-          arbiter: {}
-          encryption:
-            kms: {}
-          externalStorage: {}
-          managedResources:
-            cephBlockPools: {}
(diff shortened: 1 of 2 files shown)


dhaiducek (approved):

policygenerator/policy-sets/stable/openshift-plus/input-odf/policy-odf-cluster.yaml
```diff
@@ -31,26 +31,37 @@ spec:
...
+            noobaa-core: {}
+            noobaa-db: {}
+            noobaa-endpoint:
+              limits:
+                cpu: 1
+                memory: 500Mi
+              requests:
+                cpu: 1
+                memory: 500Mi
+            rgw: {}
```
dhaiducek, line 47: There's definitely more configuration here than the step registry, but I'll assume that's desired here. 🙂

PR #536, opened 2025-12-16
Dynamic AMI and instance type lookup from existing MachineSets with fallback to ConfigMap values

tanfengshuang:
1. Dynamic AMI Lookup

  - Queries existing MachineSets in openshift-machine-api namespace
  - Extracts AMI ID from the first MachineSet's providerSpec.value.ami.id
  - Falls back to region+version-based ConfigMap lookup if no MachineSets exist
  - Ensures storage nodes use the same AMI as compute nodes (matching architecture)

  2. Dynamic Instance Type Lookup

  - Extracts instance type from existing MachineSets
  - Automatically handles bare metal by replacing .metal with .2xlarge
  - Falls back to awsInstanceType ConfigMap value if no MachineSets found
  - Ensures architecture compatibility between AMI and instance type

### DIFF
--- a/policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
+++ b/policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
@@ -52,7 +52,18 @@ spec:
               providerSpec:
                 value:
                   ami:
-                    id: '{{ fromConfigMap "policies" "opp-settings" (printf "%s-%s" (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.platformStatus.aws.region (fromClusterClaim "openshiftversion-major-minor")) }}'
+                    id: '{{- $dynamicAMI := "" -}}
+                         {{- $machinesets := (lookup "machine.openshift.io/v1beta1" "MachineSet" "openshift-machine-api" "") -}}
+                         {{- if $machinesets.items -}}
+                           {{- if gt (len $machinesets.items) 0 -}}
+                             {{- $dynamicAMI = (index $machinesets.items 0).spec.template.spec.providerSpec.value.ami.id -}}
+                           {{- end -}}
+                         {{- end -}}
+                         {{- if $dynamicAMI -}}
+                           {{ $dynamicAMI }}
+                         {{- else -}}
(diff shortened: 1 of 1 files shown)


policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
```diff
@@ -52,7 +52,18 @@ spec:
...
+                         {{- if $machinesets.items -}}
+                           {{- if gt (len $machinesets.items) 0 -}}
+                             {{- $dynamicAMI = (index $machinesets.items 0).spec.template.spec.providerSpec.value.ami.id -}}
+                           {{- end -}}
+                         {{- end -}}
+                         {{- if $dynamicAMI -}}
+                           {{ $dynamicAMI }}
+                         {{- else -}}
+                           {{ fromConfigMap "policies" "opp-settings" (printf "%s-%s" (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.platformStatus.aws.region (fromClusterClaim "openshiftversion-major-minor")) }}
+                         {{- end -}}'
```
dhaiducek, line 66: (Optional) Leveraging `range` directly could be a bit cleaner (and I think would cover the conditionals that are in place?):

```yaml
id: >-
  {{ $dynamicAMI := "" -}}
  {{ range $machineset := (lookup "machine.openshift.io/v1beta1" "MachineSet" "openshift-machine-api" "").items -}}
    {{ $dynamicAMI = $machineset.spec.template.spec.providerSpec.value.ami.id -}}
    {{ break -}}
  {{ end -}}
  {{ defaultAMI := fromConfigMap "policies" "opp-settings" (printf "%s-%s" (lookup "config.openshift.io/v1" "Infrastructure" "" "cluster").status.platformStatus.aws.region (fromClusterClaim "openshiftversion-major-minor")) }}
  {{ default $defaultAMI $dynamicAMI -}}
```

dhaiducek (commented):
Thanks for the update @tanfengshuang! I'm not sure I'm familiar enough with this Policy to approve it, but I have a minor comment to consider!

policygenerator/policy-sets/community/openshift-plus-setup/machine-sets.yaml
gparvin: Yes, we can improve this as we move forward.  A lot of testing has been completed with the current templates.

gparvin (approved):

PR #540, opened 2026-02-12
Add policy for kube-apiserver name clash issue

dkobras:
An extraneous entry in the default ClusterRoleBinding for the kube-apiserver created during OpenShift cluster bootstrap could be abused to grant elevated privileges to a regular user with a matching name. The legacy entry is no longer required and should be removed.

Reference: <https://access.redhat.com/solutions/7136094>


dkobras:
Note: During testing, I kept hitting the 62 character limit for policy name+namespace, and therefore opted to omit the `policy-` prefix, deviating from the naming convention used by other policies in the same folder. Feel free to adjust to your preference, of course.

ch-stark:
@dhaiducek can you please review

dhaiducek (approved):
Hi @dkobras! Thanks for the contribution! This policy is a great addition to the collection here!

dhaiducek:
...and my sincerest apologies for the delay in reviewing it! Thanks for your patience!

PR #548, opened 2026-04-14
refactor: remove duplicate Kyverno policies

jan-law:
Kyverno maintains a library of nearly-identical policies at https://kyverno.io/policies/. Remove the duplicates and edit readme to point to Kyverno's policy samples.

ref: https://redhat.atlassian.net/browse/ACM-15886

What's considered a duplicate?
- Policy is a `community/` policy and is
  - identical to Kyverno's policy sample 
    OR
  - The deleted policy is a ConfigurationPolicy wrapping an identical ClusterPolicy. User can run PolicyGenerator with the Kyverno examples to achieve the same output
    OR
  - The only difference is background evaluation enabled `spec.background = true` vs `spec.background = false` 
    OR
  - The policy name, description and metadata is identical. The validation rule looks more up-to-date in [kyverno/policies](https://github.com/kyverno/policies) than the equivalent rule in the `policy-collection` policy
- Policies are part of a policy-set for which Kyverno has an equivalent set, grouped by tag (e.g. "best-practices", "EKS Best Practices")
- If a policy-set contains policies that are unique to OCM policy-collection, duplicates within the policy-set will NOT be removed

> **Note**
> Rough table of cluster policies and their Kyverno equivalents: https://gist.github.com/jan-law/b430c2717f0d411141c4190791604156

## Summary by CodeRabbit

* **Chores**
  * Removed Kyverno policy examples and templates from community configuration management collection
  * Deleted best practices policy set documentation and related configuration files


policygenerator/kustomize/policyGenerator.yaml
```diff
@@ -33,6 +33,3 @@ policies:
     consolidateManifests: false
     manifests:
       - path: policy2_gatekeeper/gatekeeper.yaml
-  - name: policy-kyverno-require-labels
-    manifests:
-      - path: policy3_kyverno/kyverno.yaml
```
dhaiducek, line 38: This one is also referenced in the README at `policygenerator/` and is meant to serve as a sample of the Policy generator. Not sure if you want to either find a suitable replacement or remove the README reference.

policygenerator/policy-sets/community/kyverno/best-practises-for-apps/policyGenerator.yaml
```diff
@@ -19,46 +19,14 @@ policyDefaults:
   standards:
     - NIST SP 800-53
 policies:
-  - name: require-probes
+  - name: require-deployments-rolling-strategy
```
dhaiducek: I'm wondering how much value this has if there's only two policies here. Maybe it just needs to be removed entirely? Either that or we just remove the policies from `community/` and leave the generator sets in place?

policygenerator/kustomize/policyGenerator.yaml
jan-law: I'll put the same example back

policygenerator/policy-sets/community/kyverno/best-practises-for-apps/policyGenerator.yaml
jan-law: Removing the `best-practises-for-apps` folder makes sense to me. Kyverno's collection lets you search policies by selecting the "best practices" tags.

policygenerator/policy-sets/community/kyverno/security/policyGenerator.yaml
```diff
@@ -18,106 +18,6 @@ policyDefaults:
...
-    manifests:
-      - path: input/restrict-wildcard-verbs
-  - name: networking-block-nodeport-services
-    categories:
-      - SC System and Communications Protection
-    controls:
-      - SC-7 Boundary Protection
-    disabled: false
-    manifests:
-      - path: input/networking/block-nodeport-services
```
dhaiducek, line 120: Similar situation here. Even though these are duplicates, I think we might either leave them in place for now since it could make sense to keep the set, or remove the set entirely since removing the duplicated policies might dilute the value of the set (I haven't really looked at the content of what changed 😬).

(This next part got way longer than I intended, so feel free to ignore my rambling 😂 )

---

Eventually, we could leverage the Kyverno Helm chart to replace this entire set and provide this Kustomize file in its place along with a link to the Kyverno docs. I'd say we shouldn't do this in this PR because it dumps all of the results into a single Policy rather than spreading them across a PolicySet. What would be nice in this case would be to have a feature in the generator where it makes specifying `policies[].name` optional and derives the name from the `metadata.name` it's wrapping, and additionally the ability to separate these files into separate Policies so that they can be listed out in a PolicySet. Kind of like the `consolidateManifests` boolean, but separated by Policy instead of ConfigurationPolicy.

This in a `kustomization.yaml` in the `input/` directory:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmCharts:
- name: kyverno-policies
  releaseName: kyverno-policies
  repo: https://kyverno.github.io/kyverno/
  kubeVersion: "1.25.0"
  version: 3.7.1
```

And then the policies would just be:
```yaml
policies:
  - name: kyverno-security-policyset
    manifests:
      - path: kyverno-policies/
```

refs:
- https://artifacthub.io/packages/helm/kyverno/kyverno-policies
- https://kyverno.io/docs/installation/installation/#pod-security-policies-optional

jan-law: Good point. There are some SCC rules in this policy-set that aren't found in Kyverno's pod security policies. Since this security policy-set is a superset of Kyverno's, I will restore the duplicates.

Same for the other policy-set `multitenancy`. It would be useful to keep these supersets even if there are duplicates.

dhaiducek (approved):
LGTM! Thanks for the cleanup!

